PR #6, opened 2019-06-13
feat(typing): Adding a module to assist with type checking

chriswmackey:
This adds the typing module to honeybee core as we discussed @mostaphaRoudsari .  I also set the name of the honeybee Face to use the type checking such that it is always valid for both E+ and Radiance.

Note that, with this implementation, an error will be raised for all honeybee face names that exceed 100 characters after stripping out illegal characters.  I don't know if we should be adding an extra setter for `name_original` so that we can set this to be the crazy-long Revit name if we so desire. Just let me know if you think we should have a separate setter and I can add it.

### DIFF
--- a/honeybee/face.py
+++ b/honeybee/face.py
@@ -2,6 +2,7 @@
 """Honeybee Face."""
 from .properties import Properties
 from .facetype import get_type_from_normal
+from .typing import valid_string
 import honeybee.writer as writer
 from ladybug_geometry.geometry3d.face import Face3D
 from ladybug_geometry.geometry3d.pointvector import Point3D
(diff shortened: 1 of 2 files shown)


chriswmackey:
BTW, I have become more confident since our call that this is the correct place for a type checking module.  I realized that so much of the type-checking in ladybug happens on the level of the DataCollection that it really is a different animal with different type-checking needs.

mostaphaRoudsari:
Thank. Looks good! The current setter already works as the setter for `name_original` too so I don't think we need another one. That can make inconsistency between the two names.

chriswmackey:
Sounds good. Thanks!

PR #10, opened 2019-07-01
feat(objects): Adding objects for zone, aperture, and door

### LINKED ISSUE #7: Add support for aperture
This is required for `honeybee-energy` to move forward.

### LINKED ISSUE #8: Add Room class
@chriswmackey, [there is an implementation in honeybee already for room which is a good point to start](https://github.com/ladybug-tools/honeybee/blob/3a3f6e8a80c8f07ece0354942231138333f967dd/honeybee/room.py) but it needs several improvements. For instance it should be initiated from a list of Faces and not by length, width and height as is it right now. This is what I had in mind for rewriting it:

```python

class Room(object):

    def __init__(self, faces, *args):
        """create from faces."""
        # ...

    @classmethod
    def from_plan_vertices(self, vertices, height, *args):
        """Create room from plan which is represented by a list of vertices."""
        # ...

    @classmethod
    def from_width_length(self, width, length, height, *args):
        """Create a box room."""
        # ...
```

chriswmackey:
This PR is fleshing out the key objects that make up honeybee-core and lays down a lot of rules for how they interact with one another. I am only submitting this as a draft PR since I don't think it's worthwhile to write all of the tests until @mostaphaRoudsari has gotten the chance to review it (I am anticipating some feedback that will result in changes).

Over the course of fleshing everything out, I came to a couple of realizations/limitations of using `weakref.proxy` that have some important implications for the core library (assuming we want to continue using it):
1) Because all of the variables that exist within a given method on a class are garbage collected after the method is run (except for the returned variables), we will not be able to implement methods that automatically generate AND add sub-faces to a honeybee face.  We can still have methods to generate sub-faces from a base face and we can have another method that adds sub-faces to the face but these two methods have to be separate from one another such that the generated sub-faces get returned from the method and are not garbage collected.  So, if you want to add apertures to a face based on a ratio, you need to call two separate methods as you see below.
It is not the most intuitive design to make 2 calls here and I had forgotten a few times in the tests that I wrote to add the apertures to the face after generating them.  However, I think that I can personally live with making two calls instead of one for this situation if you think the benefit of weakref outweighs the lack of intuitiveness.
```
apertures = face.apertures_by_ratio(0.4)
face.add_apertures(apertures)
```
2) One thing that I would not be able to live with, however, is the use of weakref for the relationship between faces and zones.  This is partly because of the fact noted above, in that it prevents us from having classmethods on the zone that allow us to create a zone without having honeybee faces to begin with (something that I think we agree is critical for if you want to make a zone from a ladybug_geometry Polyface3D or create a zone from a box).  However, it's also because the zone relies on the faces to define its geometry in a way that none of the other objects in the library rely on one another.  So, if we want to have properties on the zone like `volume` or we want to have the zone automatically flip face geometries so that they are all pointing outward from the zone, we really can't use weakrefs because the zone really can't exist in a geometric sense without the faces.

Among a few other things to note that I would really like your feedback on @mostaphaRoudsari :
1) You will see that methods to generate shade geometries from apertures are implemented on the Face level of the schema and not the Aperture (where I initially tried to put them and where we would probably agree that they make the most sense).  The reason for this is that, because shades are on the Face level and not the sub-face level, you get a circular import when you try to `import Face` in `aperture.py`.  I wasn't sure if you preferred to deal with this via importing Face within the shade method or doing something in `__init__.py` like we do with datacollections and datatype so I left it as is for now for you to comment.
2) The implementation of `apreturetypes` is in line with how we are implementing `ConstructionSets` in `honeybee-energy` and I wanted to know if you feel that this is ok.  We could also try to deal with it in the same way that we deal with adiabatic boundary conditions in that it exists in `honeybee_energy` and extends base window types but there would probably be no base window types if this was the case. There was no need for a `doortypes`, since all of the different types of doors in the ConstructionSet are defined by their boundary condition or relationship to parent.
3) I just wanted to note that all of the capabilities of the [honeybee room](https://github.com/ladybug-tools/honeybee/blob/master/honeybee/room.py) have been implemented on the Zone. When you initialize the Zone `from_box`, you get something that is virtually identical to the room except that you have multiple options for adding apertures (including `apertures_by_ratio`).  When you generate a mesh grid from the zone, you will get the grid oriented along the width/depth dimensions of the box.  The one thing that you will need to add to give it all of the room capabilities is the method that auto-generates the Radiance view, which I imagine you will extend the zone with.
4) One thing that I initially tried to implement in the core was enforcing checks that apertures were valid sub-faces of their parent face (in the same plane and completely bounded by it) but I realized this is very difficult if we want to make "the pizza" that can accept geometry from all different models of different tolerances.  A similar situation happened when I tried enforce closed zone geometry in the `__init__` of the Zone in that it virtually kills the ability to import any model that doesn't have its tolerance known or is intentionally not closed.  All of the tools exist within `ladybug_geometry' to perform these geometry checks and, for the case of the zone, I implemented such checking abilities on all of the classmethods.  However, it seems that we might ultimately need to run these checks on the plugin end rather than make them a universally-enforced part of the core.  Let me know if you can live with this.
5) Generally, there are a lot of rules that I implemented across the setters in order to ensure that we don't get things like apertures being assigned to shade geometries or air walls.  This should generally mean that people will get an error when the make something incorrect rather than waiting for the engine to give them an error, which is what we want.  However, I was wondering if you thought some of the setters might be too bogged down with checks that we should consider removing them.

resolves #8 
resolves #7

### DIFF
--- a/honeybee/_lockable.py
+++ b/honeybee/_lockable.py
@@ -1,0 +1,96 @@
+# coding=utf-8
+"""Descriptor that for locking and unlocking an object, preventing attrbute setting."""
+from functools import wraps
+from inspect import getmro
+
+
+def lockable(cls):
+    """A decorator for making lockable class.
+
+    If a class is locked, attributes can no longer be set on the class.
+    The class will throw an AttributeError if one tries to do so.
+
+    Classes start out as unlocked but can be locked by calling the `lock()` method.
+    If the class needs to be unlocked again, calling the `unlock()` method will
+    unlock it and allow you to edit attributes on the class.  You can also create
+    a class that starts out as locked by putting `self._locked = True` at the end
+    of a class's `__init__` method.
+
+    Note that classes using __slots__ must specify a '_locked' variable within
+    the __slots__ in order to work correctly with this decorator.
+
+    This example if modified from:
+    http://stackoverflow.com/questions/3603502/prevent-creating-new-attributes-outside-init
+
+    Usage:
+        @lockable
+        class Foo(object):
+            def __init__(self):
+                self.bar = 10
+
+        foo = Foo()
+        foo.bar = 20
+
+        foo.lock()
+        try:
+            foo.bar = 30
+        except AttributeError as e:
+            print(e)
+        > Failed to set bar to 30. bar cannot be set on Foo while it is locked.
+          The unlock() method can unlock the class but you do so at your own risk.
+          Objects are typically locked when they are referenced from many other objects.
+          So it is usually better to make a new instance of Foo and proceed using that.
+
+        foo.unlock()
+        foo.bar = 30
+        foo.lock()
+    """
+
+    def lockedsetattr(self, key, value):
+        """Method to overwrite __setattr__ on the decorated class."""
+        if hasattr(self, '_locked') and self._locked and not key == '_locked':
+            raise AttributeError(
+                'Failed to set {1} to {2}. {1} cannot be set on {0} while it is locked.'
+                '\nThe unlock() method can unlock the class but you do so at '
+                'your own risk.\nObjects are typically locked when they are referenced '
+                'from several other objects.\n So it is usually better practice to'
+                'make a new instance of {0} and proceed using that.'.format(
+                    cls.__name__, key, value))
+        else:
+            object.__setattr__(self, key, value)
+
+    def init_decorator(func):
+        """Initialize the locabke decorator for the class."""
+
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            func(self, *args, **kwargs)
+        return wrapper
+
+    def lock(self):
+        self._locked = True
+
+    def unlock(self):
+        self._locked = False
+
+    # overwrite the various methods on the class to support lockability
+    cls.__setattr__ = lockedsetattr
+    cls.__init__ = init_decorator(cls.__init__)
+    if not hasattr(cls, 'lock'):  # allow developers to add their own lock method
+        cls.lock = lock
+    if not hasattr(cls, 'unlock'):  # allow developers to add their own unlock method
+        cls.unlock = unlock
+
+    # if the class uses __slots__, check that _locked is somewhere in inheritance tree
+    if hasattr(cls, '__slots__'):
+        _all_good = False
+        for parent_class in getmro(cls)[:-1]:
+            if '_locked' in parent_class.__slots__:
+                _all_good = True
+                break
+        if not _all_good:
+            raise AttributeError(
+                'When using the @locakble decorator on a class with __slots__, '
+                'a "_locked" variable must be specified within __slots__.')
+
+    return cls
--- a/honeybee/aperture.py
+++ b/honeybee/aperture.py
@@ -1,0 +1,535 @@
+# coding: utf-8
+"""Honeybee Aperture."""
+from .properties import ApertureProperties
+from .aperturetype import aperture_types
+from .boundarycondition import boundary_conditions, Outdoors, Surface
+from .shade import Shade
+from .typing import valid_string
+import honeybee.writer as writer
+
+from ladybug_geometry.geometry3d.pointvector import Point3D, Vector3D
+from ladybug_geometry.geometry3d.plane import Plane
+from ladybug_geometry.geometry3d.face import Face3D
+
+import math
+
+
+class Aperture(object):
+    """A single planar aperture face.
+
+    Properties:
+        name
+        name_original
+        geometry
+        vertices
+        upper_left_vertices
+        triangulated_mesh3d
+        normal
+        center
+        area
+        perimeter
+        type
+        boundary_condition
+        parent
+        has_parent
+    """
+    TYPES = aperture_types
+    __slots__ = ('_name', '_name_original', '_geometry', '_parent',
+                 '_type', '_boundary_condition', '_properties')
+
+    def __init__(self, name, geometry, type=None, boundary_condition=None):
+        """A single planar aperture in a face.
+
+        Args:
+            name: Aperture name. Must be < 100 characters.
+            geometry: A ladybug-geometry Face3D.
+            type: Aperture type. Default: Window
+            boundary_condition: Boundary condition object (Outdoors, Surface).
+                Default: Outdoors.
+        """
+        # process the name
+        self._name = valid_string(name, 'honeybee aperture name')
+        self._name_original = name
+
+        # process the geometry
+        assert isinstance(geometry, Face3D), \
+            'Expected ladybug_geometry Face3D. Got {}'.format(type(geometry))
+        self._geometry = geometry
+        self._parent = None  # _parent will be set when the Aperture is added to a Face
+
+        # process the type and boundary condition
+        self.type = type or aperture_types.window
+        self.boundary_condition = boundary_condition or boundary_conditions.outdoors
+
+        # initialize properties for extensions
+        self._properties = ApertureProperties(self)
+
+    @classmethod
+    def from_vertices(cls, name, vertices, type=None, boundary_condition=None):
+        """Create an Aperture from vertices with each vertex as an iterable of 3 floats.
+
+        Args:
+            name: Aperture name. Must be < 100 characters.
+            vertices: A flattened list of 3 or more vertices as (x, y, z).
+            type: Aperture type. Default: Window
+            boundary_condition: Boundary condition object (eg. Outdoors, Surface).
+                Default: Outdoors.
+        """
+        geometry = Face3D(tuple(Point3D(*v) for v in vertices))
+        return cls(name, geometry, type, boundary_condition)
+
+    @property
+    def name(self):
+        """The aperture name (including only legal characters)."""
+        return self._name
+
+    @property
+    def name_original(self):
+        """Original input name by user.
+
+        If there are no illegal characters in name then name and name_original will
+        be the same. Legal characters are ., A-Z, a-z, 0-9, _ and -.
+        Invalid characters are automatically removed from the original name for
+        compatability with simulation engines.
+        """
+        return self._name_original
+
+    @property
+    def geometry(self):
+        """A ladybug_geometry Face3D object representing the aperture."""
+        return self._geometry
+
+    @property
+    def vertices(self):
+        """List of vertices for the aperture (in counter-clockwise order)."""
+        return self._geometry.vertices
+
+    @property
+    def upper_left_vertices(self):
+        """List of vertices starting from the upper-left corner.
+
+        This property should be used when exporting to EnergyPlus / OpenStudio.
+        """
+        return self._geometry.upper_left_counter_clockwise_vertices
+
+    @property
+    def triangulated_mesh3d(self):
+        """A ladybug_geometry Mesh3D of the aperture geometry composed of triangles.
+
+        In EnergyPlus / OpenStudio workflows, this property is used to subdivide
+        the aperture when it has more than 4 vertices. This is necessary since
+        EnergyPlus cannot accept sub-faces with more than 4 vertices.
+        """
+        return self._geometry.triangulated_mesh3d
+
+    @property
+    def normal(self):
+        """A ladybug_geometry Vector3D for the direction the aperture is pointing.
+        """
+        return self._geometry.normal
+
+    @property
+    def center(self):
+        """A ladybug_geometry Point3D for the center of the aperture.
+
+        Note that this is the center of the bounding rectangle around this geometry
+        and not the area centroid.
+        """
+        return self._geometry.center
+
+    @property
+    def area(self):
+        """The area of the aperture."""
+        return self._geometry.area
+
+    @property
+    def perimeter(self):
+        """The perimeter of the aperture."""
+        return self._geometry.perimeter
+
+    @property
+    def type(self):
+        """Object for Type of Aperture (ie. Window)."""
+        return self._type
+
+    @type.setter
+    def type(self, value):
+        assert value in self.TYPES, '{} is not a valid face type.'.format(value)
+        self._type = value
+
+    @property
+    def boundary_condition(self):
+        """Boundary condition of this aperture."""
+        return self._boundary_condition
+
+    @boundary_condition.setter
+    def boundary_condition(self, value):
+        assert isinstance(value, (Outdoors, Surface)), \
+            'Aperture only supports Outdoor or Surface boundary condition. ' \
+            'Got {}'.format(type(value))
+        self._boundary_condition = value
+
+    @property
+    def parent(self):
+        """Parent Face if assigned. None if not assigned."""
+        return self._parent
+
+    @property
+    def has_parent(self):
+        """Boolean noting whether this Aperture has a parent Face."""
+        return self._parent is not None
+
+    def set_adjacency(self, other_aperture):
+        """Set this aperture to be adjacent to another.
+
+        Note that this method does not verify whether the other_aperture geometry is
+        co-planar or compatible with this one so it is recommended that a test
+        be performed before using this method in order to verify these criteria.
+        The Face3D.is_centered_adjacent() or the Face3D.is_geometrically_equivalent()
+        methods are both suitable for this purpose.
+
+        Args:
+            other_aperture: Another Aperture object to be set adjacent to this one.
+        """
+        assert isinstance(other_aperture, Aperture), \
+            'Expected Aperture. Got {}.'.format(type(other_aperture))
+        self.boundary_condition = Surface(other_aperture)
+        other_aperture.boundary_condition = Surface(self)
+
+    def overhang(self, depth, angle=0, indoor=False, tolerance=0):
+        """Get a single overhang for this Aperture.
+
+        Args:
+            depth: A number for the overhang depth.
+            angle: A number for the for an angle to rotate the overhang in degrees.
+                Default is 0 for no rotation.
+            indoor: Boolean for whether the overhang should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+            tolerance: An optional value to return None if the overhang has a length less
+                than the tolerance. Default is 0, which will always yeild an overhang.
+
+        Returns:
+            overhang: A Shade object for the Aperture overhang. You may want to run
+                the check_non_zero() method on the resulting Shade to verify
+                the geometry is subtantial when the Aperture does not have a flat top.
+        """
+        overhang = self.louvers_by_number(
+            1, depth, angle=angle, indoor=indoor, tolerance=tolerance)
+        return overhang[0] if len(overhang) != 0 else None
+
+    def right_fin(self, depth, angle=0, indoor=False, tolerance=0):
+        """Get a single vertical fin on the right side of this Aperture.
+
+        Args:
+            depth: A number for the fin depth.
+            angle: A number for the for an angle to rotate the fin in degrees.
+                Default is 0 for no rotation.
+            indoor: Boolean for whether the fin should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+            tolerance: An optional value to return None if the fin has a length less
+                than the tolerance. Default is 0, which will always yeild a fin.
+
+        Returns:
+            fin: A Shade object for the Aperture fin. You may want to run
+                the check_non_zero() method on the resulting Shade to verify
+                the geometry is subtantial when the Aperture does not have a
+                flat right side.
+        """
+        fin = self.louvers_by_number(
+            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
+            indoor=indoor, tolerance=tolerance)
+        return fin[0] if len(fin) != 0 else None
+
+    def left_fin(self, depth, angle=0, indoor=False, tolerance=0):
+        """Get a single vertical fin on the left side of this Aperture.
+
+        Args:
+            depth: A number for the fin depth.
+            angle: A number for the for an angle to rotate the fin in degrees.
+                Default is 0 for no rotation.
+            indoor: Boolean for whether the fin should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+            tolerance: An optional value to return None if the fin has a length less
+                than the tolerance. Default is 0, which will always yeild a fin.
+
+        Returns:
+            fin: A Shade object for the Aperture fin. You may want to run
+                the check_non_zero() method on the resulting Shade to verify
+                the geometry is subtantial when the Aperture does not have a
+                flat right side.
+        """
+        fin = self.louvers_by_number(
+            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
+            flip_start_side=True, indoor=indoor, tolerance=tolerance)
+        return fin[0] if len(fin) != 0 else None
+
+    def extruded_border(self, depth, indoor=False):
+        """Get a list of Shade objects from the extruded border of this aperture.
+
+        Args:
+            depth: A number for the extrusion depth.
+            indoor: Boolean for whether the extrusion should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+
+        Returns:
+            extrusion: A list of Shade objects for the extruded border.
+        """
+        extru_vec = self.normal if indoor is False else self.normal.reverse()
+        extru_vec = extru_vec * depth
+        extrusion = []
+        shd_count = 0
+        for seg in self.geometry.boundary_segments:
+            shade_geo = Face3D.from_extrusion(seg, extru_vec)
+            extrusion.append(
+                Shade('{}_Shd{}'.format(self.name_original, shd_count), shade_geo))
+            shd_count += 1
+        if self.geometry.has_holes:
+            for hole in self.geometry.hole_segments:
+                for seg in hole:
+                    shade_geo = Face3D.from_extrusion(seg, extru_vec)
+                    extrusion.append(
+                        Shade('{}_Shd{}'.format(self.name_original, shd_count),
+                              shade_geo))
+                    shd_count += 1
+        return extrusion
+
+    def louvers_by_number(self, louver_count, depth, offset=0, angle=0,
+                          contour_vector=Vector3D(0, 0, 1), flip_start_side=False,
+                          indoor=False, tolerance=0):
+        """Get a list of louvered Shade objects covering this Aperture.
+
+        Args:
+            louver_count: A positive integer for the number of louvers to generate.
+            depth: A number for the depth to extrude the louvers.
+            offset: A number for the distance to louvers from this aperture.
+                Default is 0 for no offset.
+            angle: A number for the for an angle to rotate the louvers in degrees.
+                Default is 0 for no rotation.
+            contour_vector: A Vector3D for the direction along which contours
+                are generated. Default is Z-Axis, which generates horizontal louvers.
+            flip_start_side: Boolean to note whether the side the louvers start from
+                should be flipped. Default is False to have louvers on top or right.
+                Setting to True will start contours on the bottom or left.
+            indoor: Boolean for whether louvers should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+            tolerance: An optional value to remove any louvers with a length less
+                than the tolerance. Default is 0, which will include all louvers
+                no matter how small.
+
+        Returns:
+            louvers: A list of Shade objects that cover the Aperture.
+        """
+        assert louver_count > 0, 'louver_count must be greater than 0.'
+        angle = math.radians(angle)
+        louvers = []
+        ap_geo = self.geometry if indoor is False else self.geometry.flip()
+        shade_faces = ap_geo.countour_fins_by_number(
+            louver_count, depth, offset, angle,
+            contour_vector, flip_start_side, tolerance)
+        for i, shade_geo in enumerate(shade_faces):
+            louvers.append(Shade('{}_Shd{}'.format(self.name_original, i), shade_geo))
+        return louvers
+
+    def louvers_by_distance_between(self, distance, depth, offset=0, angle=0,
+                                    contour_vector=Vector3D(0, 0, 1),
+                                    flip_start_side=False, indoor=False, tolerance=0):
+        """Get a list of louvered Shade objects covering this Aperture.
+
+        Args:
+            distance: A number for the approximate distance between each louver.
+            depth: A number for the depth to extrude the louvers.
+            offset: A number for the distance to louvers from this aperture.
+                Default is 0 for no offset.
+            angle: A number for the for an angle to rotate the louvers in degrees.
+                Default is 0 for no rotation.
+            contour_vector: A Vector3D for the direction along which contours
+                are generated. Default is Z-Axis, which generates horizontal louvers.
+            flip_start_side: Boolean to note whether the side the louvers start from
+                should be flipped. Default is False to have contours on top or right.
+                Setting to True will start contours on the bottom or left.
+            indoor: Boolean for whether louvers should be generated facing the
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+            tolerance: An optional value to remove any louvers with a length less
+                than the tolerance. Default is 0, which will include all louvers
+                no matter how small.
+
+        Returns:
+            louvers: A list of Shade objects that cover the Aperture.
+        """
+        angle = math.radians(angle)
+        louvers = []
+        ap_geo = self.geometry if indoor is False else self.geometry.flip()
+        shade_faces = ap_geo.countour_fins_by_distance_between(
+            distance, depth, offset, angle,
+            contour_vector, flip_start_side, tolerance)
+        for i, shade_geo in enumerate(shade_faces):
+            louvers.append(Shade('{}_Shd{}'.format(self.name_original, i), shade_geo))
+        return louvers
+
+    def move(self, moving_vec):
+        """Move this Aperture along a vector.
+
+        Args:
+            moving_vec: A ladybug_geometry Vector3D with the direction and distance
+                to move the face.
+        """
+        self._geometry = self.geometry.move(moving_vec)
+
+    def rotate(self, axis, angle, origin):
+        """Rotate this Aperture by a certain angle around an axis and origin.
+
+        Args:
+            axis: A ladybug_geometry Vector3D axis representing the axis of rotation.
+            angle: An angle for rotation in degrees.
+            origin: A ladybug_geometry Point3D for the origin around which the
+                object will be rotated.
+        """
+        self._geometry = self.geometry.rotate(axis, math.radians(angle), origin)
+
+    def rotate_xy(self, angle, origin):
+        """Rotate this Aperture counterclockwise in the world XY plane by a certain angle.
+
+        Args:
+            angle: An angle in degrees.
+            origin: A ladybug_geometry Point3D for the origin around which the
+                object will be rotated.
+        """
+        self._geometry = self.geometry.rotate_xy(math.radians(angle), origin)
+
+    def reflect(self, plane):
+        """Reflect this Aperture across a plane with the input normal vector and origin.
+
+        Args:
+            plane: A ladybug_geometry Plane across which the object will
+                be reflected.
+        """
+        assert isinstance(plane, Plane), \
+            'Expected ladybug_geometry Plane. Got {}.'.format(type(plane))
+        self._geometry = self.geometry.reflect(plane.n, plane.o)
+
+    def scale(self, factor, origin=None):
+        """Scale this Aperture by a factor from an origin point.
+
+        Args:
+            factor: A number representing how much the object should be scaled.
+            origin: A ladybug_geometry Point3D representing the origin from which
+                to scale. If None, it will be scaled from the World origin (0, 0, 0).
+        """
+        self._geometry = self.geometry.scale(factor, origin)
+
+    def check_planar(self, tolerance, raise_exception=True):
+        """Check whether all of the Aperture's vertices lie within the same plane.
+
+        Args:
+            tolerance: The minimum distance between a given vertex and a the
+                object's's plane at which the vertex is said to lie in the plane.
+            raise_exception: Boolean to note whether an ValueError should be
+                raised if a vertex does not lie within the object's plane.
+        """
+        try:
+            return self.geometry.validate_planarity(tolerance, raise_exception)
+        except ValueError as e:
+            raise ValueError('Aperture "{}" is not planar.\n{}'.format(
+                self.name_original, e))
+
+    def check_self_intersecting(self, raise_exception=True):
+        """Check whether the edges of the Aperture intersect one another (like a bowtwie)
+
+        Args:
+            raise_exception: If True, a ValueError will be raised if the object
+                intersects with itself. Default: True.
+        """
+        if self.geometry.is_self_intersecting:
+            if raise_exception:
+                raise ValueError('Aperture "{}" has self-intersecting edges.'.format(
+                    self.name_original))
+            return False
+        return True
+
+    def check_non_zero(self, tolerance=0.0001, raise_exception=True):
+        """Check whether the area of the Aperture is above a certain "zero" tolerance.
+
+        Args:
+            tolerance: The minimum acceptable area of the object. Default is 0.0001,
+                which is equal to 1 cm2 when model units are meters. This is just
+                above the smalest size that OpenStudio will accept.
+            raise_exception: If True, a ValueError will be raised if the object
+                area is below the tolerance. Default: True.
+        """
+        if self.area < tolerance:
+            if raise_exception:
+                raise ValueError(
+                    'Aperture "{}" geometry is too small. Area must be at least {}. '
+                    'Got {}.'.format(self.name_original, tolerance, self.area))
+            return False
+        return True
+
+    @property
+    def properties(self):
+        """Aperture properties, including Radiance, Energy and other properties."""
+        return self._properties
+
+    @property
+    def to(self):
+        """Aperture writer object.
+
+        Use this method to access Writer class to write the aperture in other formats.
+
+        Usage:
+            aperture.to.idf(aperture) -> idf string.
+            aperture.to.radiance(aperture) -> Radiance string.
+        """
+        raise NotImplementedError('Aperture does not yet support writing to files.')
+        return writer
+
+    def to_dict(self, abridged=False, included_prop=None):
+        """Return Aperture as a dictionary.
+
+        Args:
+            abridged: Boolean to note whether the full dictionary describing the
+                object should be returned (False) or just an abridged version (True).
+                Default: False.
+            included_prop: List of properties to filter keys that must be included in
+                output dictionary. For example ['energy'] will include 'energy' key if
+                available in properties to_dict. By default all the keys will be
+                included. To exclude all the keys from plugins use an empty list.
+        """
+        base = {
+            'type': self.__class__.__name__,
+            'name': self.name,
+            'name_original': self.name_original,
+            'geometry': self._geometry.to_dict(),
+            'aperture_type': self.type.to_dict(),
+            'properties': self.properties.to_dict(abridged, included_prop)
+        }
+        if 'energy' in base['properties']:
+            base['boundary_condition'] = self.boundary_condition.to_dict(full=True)
+        else:
+            base['boundary_condition'] = self.boundary_condition.to_dict(full=False)
+        if self.parent:
+            base['parent'] = self.parent.name
+        else:
+            base['parent'] = None
+        return base
+
+    def duplicate(self):
+        """Get a copy of this object."""
+        return self.__copy__()
+
+    def ToString(self):
+        return self.__repr__()
+
+    def __copy__(self):
+        new_ap = Aperture(self.name_original, self.geometry, self.type,
+                          self.boundary_condition)
+        new_ap._properties.duplicate_extension_attr(self._properties)
+        return new_ap
+
+    def __repr__(self):
+        return 'Aperture: %s' % self.name_original
--- a/honeybee/aperturetype.py
+++ b/honeybee/aperturetype.py
@@ -1,0 +1,51 @@
+# coding=utf-8
+"""Aperture Types."""
+
+
+class _ApertureType(object):
+    __slots__ = ()
+
+    @property
+    def name(self):
+        return self.__class__.__name__
+
+    def to_dict(self):
+        """ApertureType as a dictionary."""
+        ap_type_dict = {
+            'type': self.name}
+        return ap_type_dict
+
+    def ToString(self):
+        return self.__repr__()
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return self.name
+
+
+class Window(_ApertureType):
+    """Type for static apertures that are not intended to be opened."""
+    __slots__ = ()
+    pass
+
+
+class _Types(object):
+    """Aperture types."""
+
+    def __init__(self):
+        self._window = Window()
+
+    @property
+    def window(self):
+        return self._window
+
+    def __contains__(self, value):
+        return isinstance(value, _ApertureType)
+
+
+aperture_types = _Types()
--- a/honeybee/boundarycondition.py
+++ b/honeybee/boundarycondition.py
@@ -1,125 +1,163 @@
-"""Face Boundary Condition."""
-import weakref
+"""Boundary Condition for Face, Aperture, Door."""
+from .typing import float_in_range
 
 
 class _BoundaryCondition(object):
+    """Base boundary condition class."""
 
     __slots__ = (
-        '_sun_exposure', '_wind_exposure', '_boundary_condition_object', '_view_factor'
-    )
+        '_sun_exposure', '_wind_exposure', '_view_factor')
 
-    def __init__(self):
-        self.sun_exposure = False
-        self.wind_exposure = False
-        self.boundary_condition_object = None
-        self.view_factor = 'autocalculate'
+    def __init__(self, sun_exposure=False, wind_exposure=False,
+                 view_factor='autocalculate'):
+        """Initialize Boundary condition.
+
+        Args:
+            sun_exposure: A boolean noting whether the boundary is exposed to sun.
+                Default: False.
+            wind_exposure: A boolean noting whether the boundary is exposed to wind.
+                Default: False.
+            view_factor: A number between 0 and 1 for the view factor to the ground.
+                This input can also be the word 'autocalculate' to have the view
+                factor automatically calculated.  Default: 'autocalculate'.
+        """
+        assert isinstance(sun_exposure, bool), \
+            'Input sun_exposure must be a Boolean. Got {}.'.format(type(sun_exposure))
+        self._sun_exposure = sun_exposure
+        assert isinstance(wind_exposure, bool), \
+            'Input wind_exposure must be a Boolean. Got {}.'.format(type(wind_exposure))
+        self._wind_exposure = wind_exposure
+        if view_factor == 'autocalculate':
+            self._view_factor = view_factor
+        else:
+            self._view_factor = float_in_range(
+                view_factor, 0.0, 1.0, 'view factor to ground')
 
     @property
     def name(self):
+        """Name of the boundary condition (ie. 'Outdoors', 'Ground')."""
         return self.__class__.__name__
 
     @property
     def sun_exposure(self):
+        """A boolean noting whether the boundary is exposed to sun."""
         return self._sun_exposure
-
-    @sun_exposure.setter
-    def sun_exposure(self, value):
-        assert isinstance(value, bool), \
-            'Input value must be a Boolean not {}.'.format(type(value))
-        self._sun_exposure = value
 
     @property
     def wind_exposure(self):
+        """A boolean noting whether the boundary is exposed to wind."""
         return self._wind_exposure
-
-    @wind_exposure.setter
-    def wind_exposure(self, value):
-        assert isinstance(value, bool), \
-            'Input value must be a Boolean not {}.'.format(type(value))
-        self._wind_exposure = value
-
-    @property
-    def boundary_condition_object(self):
-        return self._boundary_condition_object
-
-    @boundary_condition_object.setter
-    def boundary_condition_object(self, bco):
-        self._boundary_condition_object = bco
 
     @property
     def view_factor(self):
+        """The view factor to the ground as a number or 'autocalculate'."""
         return self._view_factor
-
-    @view_factor.setter
-    def view_factor(self, vf):
-        if vf == 'autocalculate':
-            self._view_factor = vf
-        else:
-            self._view_factor = float(vf)
 
     @property
     def sun_exposure_idf(self):
+        """Text string for sun exposure, which is write-able into an IDF."""
         return 'NoSun' if not self.wind_exposure else 'SunExposed'
 
     @property
     def wind_exposure_idf(self):
+        """Text string for wind exposure, which is write-able into an IDF."""
         return 'NoWind' if not self.wind_exposure else 'WindExposed'
-
-    @property
-    def boundary_condition_object_idf(self):
-        return '' if not self.boundary_condition_object else \
-            self.boundary_condition_object.name
 
     def to_dict(self, full=False):
         """Boundary condition as a dictionary.
-        
-        args:
-            full: Set to True to get the full dictionary which includes energy simulation
-            specific keys such as sun_exposure, wind_exposure and view_factor. Default is
-            set to False.
-        
-        returns:
-            A dictionary.
+
+        Args:
+            full: Set to True to get the full dictionary which includes energy
+                simulation specific keys such as sun_exposure, wind_exposure and
+                view_factor. Default: False.
         """
         if full:
-            return {
-                'name': self.name,
-                'bc_object': self.boundary_condition_object_idf,
+            bc_dict = {
+                'type': self.name,
                 'sun_exposure': self.sun_exposure_idf,
                 'wind_exposure': self.wind_exposure_idf,
                 'view_factor': self.view_factor
             }
         else:
-            return {
-                'name': self.name,
-                'bc_object': self.boundary_condition_object_idf
+            bc_dict = {
+                'type': self.name,
             }
+        if hasattr(self, 'boundary_condition_object'):  # Surface boundary condition
+            bc_dict['bc_object'] = self.boundary_condition_object
+        return bc_dict
+
+    def ToString(self):
+        return self.__repr__()
 
     def __repr__(self):
         return self.name
 
 
 class Surface(_BoundaryCondition):
+    """Bondary condition when an object is adjacent to another object."""
 
-    __slots__ = ()
+    __slots__ = ('_boundary_condition_objects',)
 
-    def __init__(self, other_object=None):
+    def __init__(self, other_object):
+        """Initialize Surface boundary condition.
+
+        Args:
+            other_object: Another object (Face, Aperture, Door) of the same type
+                that this boundary condition is assigned.  This other_object will be
+                set as the adjacent object in this boundary condition.
+        """
         _BoundaryCondition.__init__(self)
-        if other_object:
-            self.boundary_condition_object = weakref.proxy(other_object)
+        self._boundary_condition_objects = []
+        if other_object is not None:
+            self._boundary_condition_objects.append(other_object.name)
+            if other_object.has_parent:
+                self._boundary_condition_objects.append(other_object.parent.name)
+                if other_object.parent.has_parent:
+                    self._boundary_condition_objects.append(
+                        other_object.parent.parent.name)
+
+    @property
+    def boundary_condition_objects(self):
+        """A tuple of up to 3 object names that are adjacent to this one.
+
+        The first object is always the one that is immediately adjacet and is of
+        the same object type (Face, Aperture, Door).
+        When this boundary condition is applied to a Face, the second object in the
+        tuple will be the parent Room of the adjacent object.
+        When the boundary condition is allplied to a sub-face (Door or Aperture),
+        the second object will be the parent Face of the sub-face and the third
+        object will be the parent Room of the adjacent sub-face.
+        """
+        return tuple(self._boundary_condition_objects)
+
+    @property
+    def boundary_condition_object(self):
+        """The name of the object adjacent to this one."""
+        return self._boundary_condition_objects[0]
 
 
 class Outdoors(_BoundaryCondition):
-
+    """Outdoor boundary condition."""
     __slots__ = ()
 
-    def __init__(self):
-        _BoundaryCondition.__init__(self)
-        self.wind_exposure = True
-        self.sun_exposure = True
+    def __init__(self, sun_exposure=True, wind_exposure=True,
+                 view_factor='autocalculate'):
+        """Initialize Outdoors boundary condition.
+
+        Args:
+            sun_exposure: A boolean noting whether the boundary is exposed to sun.
+                Default: True.
+            wind_exposure: A boolean noting whether the boundary is exposed to wind.
+                Default: True.
+            view_factor: A number between 0 and 1 for the view factor to the ground.
+                This input can also be the word 'autocalculate' to have the view
+                factor automatically calculated.  Default: 'autocalculate'.
+        """
+        _BoundaryCondition.__init__(self, sun_exposure, wind_exposure, view_factor)
 
 
 class Ground(_BoundaryCondition):
+    """Ground boundary condition."""
     __slots__ = ()
     pass
 
@@ -139,7 +177,7 @@ class _BoundaryConditions(object):
     def ground(self):
         return self._ground
 
-    def surface(self, other_object=None):
+    def surface(self, other_object):
         return Surface(other_object)
 
     def __contains__(self, value):
@@ -147,3 +185,25 @@ class _BoundaryConditions(object):
 
 
 boundary_conditions = _BoundaryConditions()
+
+
+def get_bc_from_position(positions, ground_depth=0):
+    """Return a boundary condition based on the relationship to a gound plane.
+
+    Positions that are entirely at or below the ground_depth will get a Ground
+    boundary condition. If there are any positions above the ground_depth, an
+    Outdoors boundary condition will be returned.
+
+    args:
+        positions: A list of ladybug_geometry Point3D objects represeting the
+            vertices of an object.
+        ground_depth: The Z value above which positions are considered Outdoors
+            instead of Ground.
+
+    Returns:
+        Face type instance.
+    """
+    for position in positions:
+        if position.z > ground_depth:
+            return boundary_conditions.outdoors
+    return boundary_conditions.ground
--- a/honeybee/door.py
+++ b/honeybee/door.py
@@ -1,0 +1,340 @@
+# coding: utf-8
+"""Honeybee Door."""
+from .properties import DoorProperties
+from .boundarycondition import boundary_conditions, Outdoors, Surface
+from .typing import valid_string
+import honeybee.writer as writer
+
+from ladybug_geometry.geometry3d.pointvector import Point3D
+from ladybug_geometry.geometry3d.plane import Plane
+from ladybug_geometry.geometry3d.face import Face3D
+
+import math
+
+
+class Door(object):
+    """A single planar door in a honeybee Face.
+
+    Properties:
+        name
+        name_original
+        geometry
+        vertices
+        upper_left_vertices
+        triangulated_mesh3d
+        normal
+        center
+        area
+        perimeter
+        boundary_condition
+        parent
+        has_parent
+    """
+    __slots__ = ('_name', '_name_original', '_geometry', '_parent',
+                 '_boundary_condition', '_properties')
+
+    def __init__(self, name, geometry, boundary_condition=None):
+        """A single planar door in a honeybee Face.
+
+        Args:
+            name: Door name. Must be < 100 characters.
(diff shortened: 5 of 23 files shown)


mostaphaRoudsari:
Hi @chriswmackey, Thank you for posting this PR sooner than later. I think it will be easier to discuss most of this over a call but here are my quick comments.

Back to using `weakref` it should not make those limitations. I'm not sure what stops you from creating classmethods. Also not sure how garbage collection should affect your design.

Back to the rest of the notes:

1.  If you ask me they should be accessible in both levels so user can add shadings either to whole face or only to an aperture or multiple aperture. In any case, I think circular import is secondary to have the method where it is supposed to be.

2. I need to think about this more but here is my first reaction: I think keeping the energy specific ones to honeybee-energy is probably a better idea. Specially when it comes to aperture types it can confuses Radiance users that each type has a meaning when it is translated to Radiance. What I'm trying to say is if one has honeybee and honeybee-radiance should only see a base aperture type and the rest should be done in honeybee-radiance. 

3. That's great but should we call it Zone or should we call it Room in honeybee-core? I feel Room is more appropriate. Then Zone in honeybee-energy can subclass from Room in honeybee-core or just extend it. In that case, a similar approach can happen from the Radiance side as you mentioned.

4. What of we always project the points to parent Face plane before running the checks. In that case then everything will match the parent face and the tolerance will be the tolerance of projection.

5. Thank you for letting me know. I will comment on this case by case.

chriswmackey:
@mostaphaRoudsari ,

I finished adding in all of the stuff that we discussed after you posted the previous comments.

I also took a stab at addressing the issue of how to copy the objects between components when we eventually implement everything on the Grasshopper plugin side.  Rather than casting ourselves into the unknown void of `copy.deepcopy`, the `duplicate()` methods that I implemented on all mutable objects now copy only what is necessary, letting the immutable objects assigned to the base object remain unchanged between copies while calling a `duplicate` method for all children objects that are mutable.  This means that, for our energy and radiance property objects, we will have to implement a `duplicate` method on them since they will most certainly be mutable.  The `Properties` objects here in the core library are built to call these `duplicate` methods when they are, themselves, duplicated.  So this should give us a lean solution to our copying woes.

As you requested, I exposed a lot of checks that can be run (altogether, I think there are over 25 different check methods between all of the objects in this PR).  I trust people will understand that, just because we have exposed all of these checks doesn't mean that they should be run all of the time since you can easily increase the time it takes to build or import a model 100-fold by running all of them for every object. I would even go as far to say that, if people are building models from our plugins and they aren't clueless about how to set up a model properly, there shouldn't be a need to run any of them since, when people create model geometry from the plugins, we won't allow things like rooms that are not closed or apertures that are not valid sub-faces.  So I have taken a general strategy of not running most checks by default but leaving the methods there for people to check if/when they want to.  I feel that this is generally in-line with the philosophy we have used throughout the tools of not trying to hinder the expert just to catch every single mistake made by beginners.

There are two things that I did not really figure out in this PR (but really should be figured out soon in future PRs):

1) Are the energy and radiance properties actually going work as we expect them to.  We should probably run some tests with this soon but there's a huge amount of stuff to be figured out here.  To just cite a couple examples of the craziness that awaits us, I don't know how I'll handle the fact that shading faces need different energy properties than zone faces even though they are both Faces objects in the core library with the same `FaceProperty` object.  To cite another, I don't know how I am going to allow people to change the interior construction used during solve adjacency, or how to set such adjacent surfaces to adiabatic.  Will I have have to overwrite the adjacency-solving methods on the objects that I have written into the core library here?  If I overwrite them from  the energy plugin, does that mean that daylight cannot have a method to overwrite them?

2) How are we going to get the `to_dict` method of the model object to output the model schema when so many duplicate properties exist in the assigned energy and radiance properties.  I don't know where we will have the methods that consolidate things like all of the constructions across the model faces or schedules across the rooms.  Maybe you can take an initial stab at figuring this out for radiance materials @mostaphaRoudsari , since that seems like a good starting point for figuring this out.

chriswmackey:
I just wanted to add that I think I have come up with a couple of solutions to 1. but we should discuss.  

For how to deal with different properties on Shade vs. Room faces, I could have the `__init__` of the FaceEnergyProperties sense whether the face_type is a Shading and, if so, a sub-object under the face properties will be either a room face or shade face.  This will add one extra layer of complexity to access the face properties but it will work if we think the beauty of having Shadings and Room Faces as the same core object is too nice to compromise on.
So a room face construction would be accessed by:
`face_construction = face.properties.energy.room_face.construction`
And a shade face transmittance schedule would be accessed by:
`face_construction = face.properties.energy.shade_face.transmittance_schedule`
This would also mean that we would have to kill the ability to set a face to be a Shading with the face_type.setter and prevent the changing of face_type once it is a Shading since this would obviously create havoc for energy if this were changed.

I still don't have a clean solution for solve adjacency but having energy overwrite the method will work as long as you don't see anyone wanting to set an alternative rad material when they solve adjacencies.

mostaphaRoudsari:
Hi @chriswmackey, Thank you for the explanation and I agree with the approach. I will go through the code later today but here are my thoughts about your questions.

> I don't know how I'll handle the fact that shading faces need different energy properties than zone faces even though they are both Faces objects in the core library with the same FaceProperty object.

I think they should have different FaceProperty objects. Is there anything that stops us from doing that?

> I don't know how I am going to allow people to change the interior construction used during solve adjacency, or how to set such adjacent surfaces to adiabatic. Will I have have to overwrite the adjacency-solving methods on the objects that I have written into the core library here?

I see this as two separate tasks. 1. solve adjacency.  2. setting construction for adjacent faces. The first task will be done in honeybee core, the second task will be done in honeybee-energy. I would say a solution similar to what we do right now in legacy plugins should work for most of the cases. One can pass the boundary condition and construction to solve adjacency in `honeybee-energy`. `honeybee-energy` calls `honeybee-core`'s solve adjacency method and then overwrites the boundary condition / construction for interior faces.

Thinking about it more maybe we should allow passing a construction set to solve adjacency so one can set the construction for interior faces based on Face type.

> If I overwrite them from the energy plugin, does that mean that daylight cannot have a method to overwrite them?

Yes but why would one wants to set a face to adiabatic from `honeybee-radiance`? This is something that is in the scope of `honeybee-energy` library so I think it is fine to give it the full control.

 > How are we going to get the to_dict method of the model object to output the model schema when so many duplicate properties exist in the assigned energy and radiance properties. I don't know where we will have the methods that consolidate things like all of the constructions across the model faces or schedules across the rooms. Maybe you can take an initial stab at figuring this out for radiance materials @mostaphaRoudsari , since that seems like a good starting point for figuring this out.

This will happen inside the model object. Basically when you create a model from rooms and free-floating faces it compiles the input information and takes care of all the duplicate data. I can give it a try first and we can improve it further.

chriswmackey:
@mostaphaRoudsari ,

Thank you for all of the comments and I feel better about how we might handle solve adjacency and the model schema based on what you have said.

However, I have to say that forcing `Shading` to be a face_type has started to get so painful at this point (and not just from the perspective of honeybee-energy) that it is hard to describe how much of a relief it would be to have `ShadeFace` be its own object in the core.  I know that we should discuss this before I make any changes but, in preparation, I am outlining the pros of each approach below to help explain why it would be such a relief to have a `ShadeFace` object.

**Reasons for Having Separate Core Objects for Face and ShadeFace**

1. The `boundary_condition` property currently has no meaning for a `Face` that is a shading.  In it's own right, an extra meaningless property is not a deal-breaker but I can easily see someone deciding to put louvers on their building and suddenly we have thousands of copies of a meaningless property that unnecessarily eats away at memory (and BCs are not the smallest of objects given that they have all of the properties for energy simulation like `view_factor_ to_ground`). Moreover, the fact that the property is meaningless may be misleading to people interacting with the API, making them think that they have assigned a Shading to the interior of a Room by changing the boundary condition when, in fact, they should have used the `add_indoor_shading()` method on the Room.  If `ShadeFace` were its own object, we could just not have a `boundary_condition` property for it (and also not have a `face_type`), making them super-lightweight if people want to go crazy with lovers or some complex shade geometry.

2. In accordance with the note above, we currently have to run a lot of extra checks to be sure that no one adds a Shading `Face` to be a part of a `Room` volume and that no one sets a Shading face to be adjacent to a `Room` face. Having a shade be a part of a room volume clearly creates difficulties when we want to send it to energy since E+ only recognizes the 3 types (Wall, Roof, Floor). If they were separate objects, we wouldn't need these extra checks.

3. We also currently have to run checks to ensure that no one adds an `Aperture` or a `Door` to a `Face` that is a Shading.  This would be extremely easy if `ShadeFace` were its own object because we can just not put an `apertures` or a `doors` property on it.

4. This is a small point but having a separate `ShadeFace` object would solve the circular import that currently exists on `Aperture` objects in order to enable the methods that generate shades from `Aperture` geometry. If `ShadeFace` were its own object, we can separately have `Room`, `Face`, and `Aperture` import `ShadeFace`.

5. It saves us from the current hell that I have been in as I try to implement energy properties on the core `Face` object.  I was trying the workaround I noted above but it was proving to be really unsatisfying.  At the moment, I am trying to replace the `energy` attribute on the core `FaceProperties` (that is usually supposed to be an `EnergyProperty` object) with a method that is capable of returning two different types of objects depending on whether the Face it is assigned to is a Shading or a normal Face.  It feels very hacky and it also means that we will have to add more checks to the core (the ones that I mentioned in my previous post).  These checks are needed in order to stop people from turning a Shading face into a normal Face and vice versa (since such a change requires a complete change in the type of `EnergyProperties`).  Obviously, if `Face` and `ShadeFace` are separate objects, we don't have to worry about people converting between them and this whole hacky workaround gets shelved.

**Reasons for Having Shading as a FaceType**

1. It "feels" simpler, though, as the points above indicate, the reality is quite complex and we will end up running a lot of implicit checks just to get it to work.  But the "feel" of the API design is a valid thing to consider.

2. It means that we have 5 geometry objects in honeybee-core instead of 4. One extra geometry object means one extra thing that people need to learn in order to pick up the API and I don't think that we should make the decision of adding a new geometry object lightly. Just to clarify the 4 current objects are (`Room`, `Face`, `Aperture`, `Door`) and this would be changing to (`Room`, `Face`, `ShadeFace`, `Aperture`, `Door`).

**Summary**

Hopefully, that helps explain the current situation and please feel free to add to the lists if I forgot anything.  I also just wanted to note that, if we do end up having separate Python objects for `Face` and `ShadeFace`, we could still organize them in the same part of the model schema.  This would be similar to the idea we mentioned of keeping both `Apertures` and `Doors` under the same `sub_faces` part of the schema.  Also, I would be ok if we wanted to have a `HoneybeeFace` Grasshopper component that is capable of producing both `Face` and `ShadeFace` objects. However, I know that the `boundary_condition` and `energy_construction` inputs for this component are basically meaningless when someone creates a `ShadeFace` and I would worry a little about someone thinking they have changed a property of the `ShadeFace` when they input something there.

chriswmackey:
@mostaphaRoudsari ,

Now that we worked through a lot of the major issues, I think that this is finally ready for a detailed review from you.  

And, once you do the review and confirm that the general structural issues are ok, I will write tests for everything.  In the meantime, I am going to work on a related PR for `honeybee-energy`, which shows how I am thinking the core and extension libraries will talk with one another.

chriswmackey:
FYI, @mostaphaRoudsari , I just added some methods to the `Model` object that can triangulate apertures and doors for EnergyPlus when they have more than 4 vertices.  I know the use of these methods is slightly more implicit rather than explicit and I also realize that it contradicts what I said in our earlier discussion about non-planar geometry: that I want to see the exact geometry that I am simulating before I make the model.  However, I feel this case is a bit different than the non-planar geometry case for the following reasons:

1. The 4-vertex limit is a stupid limitation of EnergyPlus that really should not be there (I know EnergyPlus has built-in triangulation methods and the developers really could have taken care of this under the hood).
2. It doesn't seem right to subject the Radiance geometry to this stupid limitation.
3. Unlike curved geometry planarization to which there is a whole art, I really don't care about how such aperture/door geometry is triangulated as long as the triangles together equal the full geometry and the number of triangles is as small as possible (both of which are true for the current triangulation algorithm in `ladybug_geometry`).
4. Unlike curved geometry planarization, this type of triangulation is simple enough that we can reliably take care of it for the user without causing errors or exceptions.

The one thing that makes me a little nervous about doing this is that this means we will not be able to directly match geometry before the model-building process to results afterwards.  So I am creating a new coordination nightmare for myself when it comes to things like coloring surfaces with results and generating microclimate maps.  But I the detriment of this coordination nightmare seems to be outweighed by the benefit of taking care of this for the user.  Let me know if you agree.

mostaphaRoudsari:
It all makes sense. So I understand how do you want to deal with the change?

1. When do you see this triangulation to happen in the process? Right before writing to JSON so it goes like that to OpenStudio?

2. Does it mean that you replace the adjacent faces if exist? Is this something that one wants to run on the `Model` level when all the zones are in?

3. Is this something that OpenStudio wants to address under the hood as it is an EnergyPlus limitation which other interfaces also have to deal with.

chriswmackey:
To answer your questions:

1. Yes, I think it will happen right before writing to JSON.  It clearly needs to happen before it goes to OpenStudio but it seems like it should be in the last step right before this in order to give people the option of not triangulating if they only want to run a Radiance simulation.

2. You will see in my code and in my previous comment that " I added the methods to the `Model` object". So yes, I take care of the adjacency when it's an interior window and, yes, it needs to be on the `Model` object.

3. We can ask Dan but I am fairly confident that OpenStudio doesn't take care of it now and I'm not sure that Dan wants to add this in.

chriswmackey:
Just to further clarify point 3, I tested it and OpenStudio definitely does not handle this now.  It just ignores any Fenestration Surface with more than 4 points.  I am also just not sure that we would really want the OpenStudio SDK to automatically do this since, at the end of the day, there are going to be multiple triangles in E+ and, if we aren't controlling the process of splitting faces into triangles ourselves (including naming of them), it might be difficult for us to put the results back together afterwards.

mostaphaRoudsari:
You are right. I think it will also make people crazy if they end up with extra faces and don't know how to merge them back together.

Back to your concern about this being implicit I'm not as worried. I think you have already made it as explicit as it can be. Since the method we can expose it in case one wants to see the energy model geometry generated by honeybee before being exported to OpenStudio. If it ended up being in high demand we can have a method to only show the ones that will change and how they will change.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+
+        Returns:
+            overhang: A Shade object for the Aperture overhang. Will be None if the
+                Aperture geometry does not have a flat top.
+        """
+        overhang = self.generate_louvers_by_number(
+            1, depth, angle=angle, indoor=indoor)
+        return overhang[0] if len(overhang) != 0 else None
```
mostaphaRoudsari: Do we want to support adding overhangs, fins, etc to an aperture as shadings objects? Or are you keeping that for `honeybee-energy`? I think that will be a useful property.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+                opposite direction of the aperture normal (typically meaning
+                indoor geometry). Default: False.
+
+        Returns:
+            fin: A Shade object for the Aperture fin. Will be None if the
+                Aperture geometry does not have a flat right side.
+        """
+        fin = self.generate_louvers_by_number(
+            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0), indoor=indoor)
+        return fin[0] if len(fin) != 0 else None
```
mostaphaRoudsari: I think for case of no fin / overhang, etc we should throw a warning so the user knows that no fin has been created. This will be more important when it comes to interfaces like Grasshopper and having a warning clarifies the non flat situation.

In case you are already throwing a warning in ladybug-geometry you can ignore this comment.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        Returns:
+            fin: A Shade object for the Aperture fin. Will be None if the
+                Aperture geometry does not have a flat right side.
+        """
+        fin = self.generate_louvers_by_number(
+            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
+            flip_start_side=True, indoor=indoor)
+        return fin[0] if len(fin) != 0 else None
+
+    def extruded_border(self, depth, indoor=False):
```
mostaphaRoudsari: This will be very useful for creating daylight models from energy models. Specially now that we can get the thickness of the wall construction.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        self._geometry = self.geometry.rotate_xy(math.radians(angle), origin)
+
+    def reflect(self, plane):
+        """Reflect this Aperture across a plane with the input normal vector and origin.
+
+        Args:
+            plane: A ladybug_geometry Plane across which the object will
+                be reflected.
+        """
+        self._geometry = self.geometry.reflect(plane.n, plane.o)
```
mostaphaRoudsari: Checking for instance to be plane will be helpful here because otherwise you will get an AttributeError which is not really helpful to understand the problem.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        """Scale this Aperture by a factor from an origin point.
+
+        Args:
+            factor: A number representing how much the object should be scaled.
+            origin: A ladybug_geometry Point3D representing the origin from which
+                to scale. If None, it will be scaled from the World origin (0, 0, 0).
+        """
+        self._geometry = self.geometry.scale(factor, origin)
+
+    def is_planar(self, tolerance, raise_exception=True):
```
mostaphaRoudsari:
```suggestion
    def is_planar(self, tolerance, raise_exception=False):
```

This might be how I would use such method but by default you want get a True / False unless there is a case for checking planarity as a requirement which is different.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+                intersects with itself. Default: True.
+        """
+        if self.geometry.is_self_intersecting:
+            if raise_exception:
+                raise ValueError('Aperture "{}" has self-intersecting edges.'.format(
+                    self.name_original))
+            return False
+        return True
+
+    def is_non_zero(self, tolerance=0.0001, raise_exception=True):
```
mostaphaRoudsari: same comment here about `raise_exception`.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+
+    def duplicate(self):
+        """Get a copy of this object."""
+        return self.__copy__()
+
+    def ToString(self):
+        return self.__repr__()
+
+    def __copy__(self):
+        new_ap = Aperture(self.name_original, self.geometry, self.type.duplicate(),
```
mostaphaRoudsari: I might be wrong but I think we should not need to duplicate type and boundary condition except for Surface that boundary condition should be duplicated.

honeybee/aperturetype.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def window(self):
+        return self._window
+
+    def __contains__(self, value):
+        return isinstance(value, _ApertureType)
+
+
+aperture_types = _Types()
```
mostaphaRoudsari: I think we should go for the Singleton implementation for types so we can remove cases like this. I don't think this change is part of this PR but something that we should consider to do in the future.

honeybee/boundarycondition.py
```diff
@@ -1,116 +1,174 @@
...
             }
         else:
-            return {
-                'name': self.name,
-                'bc_object': self.boundary_condition_object_idf
+            bc_dict = {
+                'type': self.name,
             }
+        try:
+            bc_dict['bc_object'] = self.boundary_condition_object
```
mostaphaRoudsari:
```suggestion
            bc_dict['bc_object'] = self.boundary_condition_object.to_dict(full)
```

honeybee/boundarycondition.py
```diff
@@ -1,116 +1,174 @@
...
+        return self.__repr__()
+
+    def __copy__(self):
+        new_bc = self.__class__()
+        new_bc._sun_exposure = self._sun_exposure
+        new_bc._wind_exposure = self._wind_exposure
+        new_bc._view_factor = self._view_factor
+        return new_bc
+
+    def __eq__(self, other):
```
mostaphaRoudsari: Don't you need to check other properties for equality?

tests/lockable_test.py
```diff
@@ -0,0 +1,80 @@
...
+    foo = Foo()
+    foo.bar = 20
+
+    foo.lock()
+    with pytest.raises(AttributeError):
+        foo.bar = 30
+
+    foo.unlock()
+    foo.bar = 30
+    foo.lock()
```
mostaphaRoudsari: you may want to add a quick check here to ensure the value is assigned correctly when unlocked.

tests/lockable_test.py
```diff
@@ -0,0 +1,80 @@
...
+        def __init__(self):
+            self.material = Material()
+
+    constr = Construction()
+    constr.lock()
+    with pytest.raises(AttributeError):
+        constr.material = Material()
+
+    constr.material.r_value = 5
+    constr.material.lock()
```
mostaphaRoudsari: Interesting! Now I can see how overwriting `lock` and `unlock` can be helpful. In a construction for example you want to lock all the materials as well.

tests/boundary_condition_test.py
```diff
@@ -9,19 +10,16 @@
 def test_outdoors():
     bc = bcs.outdoors
     assert bc.name == 'Outdoors'
-    assert bc.sun_exposure == True
+    assert bc.sun_exposure is True
```
mostaphaRoudsari: Already sent you a link about this. `==` is not the best but is better than using `is` for checking equality for boolean values.

honeybee/typing.py
```diff
@@ -51,7 +51,8 @@ def valid_ep_string(value, input_name=''):
     and ensuring the name is not longer than 100 characters.
     """
     try:
-        val = re.sub(r'[^.\sA-Za-z0-9_-]', '', value)
+        val = ''.join(i for i in value if ord(i) < 128)  # strip out non-ascii
```
mostaphaRoudsari: This will remove white space in string. Is this what you want to happen for EP names?

honeybee/_lockable.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def lock(self):
+        self._locked = True
+
+    def unlock(self):
+        self._locked = False
+
+    cls.__setattr__ = lockedsetattr
+    cls.__init__ = init_decorator(cls.__init__)
+    if not hasattr(cls, 'lock'):  # allow developers to add their own lock method
```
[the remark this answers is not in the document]
mostaphaRoudsari: See here: https://github.com/ladybug-tools/honeybee-core/pull/10#discussion_r301805376

I can see why it is useful but I think it deserves a warning.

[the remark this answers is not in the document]
chriswmackey: Glad that you found the reason why I implemented it this way.  What type of warning were you thinking of?  It it just some extra text in the docstring?  Also bear in mind that this module is hidden and I am trying hard to make it such that none of the people who just want to script with the Python API encounter it.  It's really only something that advanced code contributors should be aware of.

As a side note, I realized that, if we want to get this decorator to work with a class that has `__slots__` people need to make sure that `'_locked'` is included in the `__slots__`.  I made a note of this in the docstring:
https://github.com/ladybug-tools/honeybee-core/pull/10/files#diff-995b5b80bda8c6cd17d271029ce81879R16

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        """
+        # process the name
+        self._name = valid_string(name, 'honeybee aperture name')
+        self._name_original = name
+
+        # process the geometry
+        assert isinstance(geometry, Face3D), \
+            'Expected ladybug_geometry Face3D. Got {}'.format(type(geometry))
+        self._geometry = geometry
+        self._parent = None  # _parent will be set when the Face is added to a Face
```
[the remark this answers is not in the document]
chriswmackey: I really wish there was a way to merge in your change here without messing up the commit history.  For now, I'll just make the same change on my end and amend the previous commit.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+
+        # process the type and boundary condition
+        self.type = type or Window()
```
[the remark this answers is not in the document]
chriswmackey: I could.  Is that better?

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        self.boundary_condition = boundary_condition or Outdoors()
```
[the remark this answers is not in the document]
chriswmackey: Ah, I didn't think of that.  Ok, I will do this in all places where we have types of face/aperture and boundary conditions that aren't "Surface".

honeybee/aperture.py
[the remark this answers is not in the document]
chriswmackey: Ok, I see why it is better.

honeybee/aperture.py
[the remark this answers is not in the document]
chriswmackey: Actually, wait. The one thing to keep in mind here is that, if someone decides to edit the `sun_exposure` or `view_factor` of the default outdoor boundary condition assigned to the Aperture, it will set all of the defaulted outdoor boundary conditions across the plugin to not have sun exposure.  Is this what we want or should we be making new instances of `Outdoors` for each aperture like I have here?  Or should we just have the boundary conditions locked whenever they get assigned to an Aperture, Face or Door.  That way, people can always make a custom boundary condition and assign it to a Face but the can't edit the assigned boundary condition directly and accidentally edit all of the default boundary conditions.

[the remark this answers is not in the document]
chriswmackey: I think the locking actually won't help us too much in the plugin because, when someone goes to duplicate a Face, the boundary condition will have to be duplicated as well unless we want to add an extra property to boundary condition in to keep track of whether the boundary condition is default.  So, after the first duplication, we will end up with tons of instances of the boundary condition anyway.
Perhaps the best solution is just to remove the `sun_exposure` and `wind_exposure` setters for the boundary condition (essentially making them immutable and thereby tying the exposure to the type of boundary condition). Along these lines, we can remove the 'view_factor` property since it seems dumb to compromise the whole usage of the boundary condition object across the library just to satisfy the one researcher who wants to set this property.  Honestly, that one researcher can just write their own measure.

What do you think?

honeybee/_lockable.py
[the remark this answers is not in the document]
mostaphaRoudsari: That is fair. You don't want every user to see a warning for lock and unlock methods being overwritten. The note in docstring should satisfy the need for developers.

honeybee/aperture.py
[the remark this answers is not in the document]
mostaphaRoudsari: I think making sun_exposure and wind_exposure and view factor read_only is a good idea. I don't think we should remove them though.

Using locked might be an overkill in this case.

One can still make a custom boundary condition from the base class if needed.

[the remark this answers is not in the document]
mostaphaRoudsari: Same comments as above. I think overwriting the setters and making them read only makes the most sense and users can create their own custom boundary condition if needed for cases like view factor.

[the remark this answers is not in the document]
chriswmackey: Sounds good and I agree.  I definitely wasn't thinking of removing the `sun_exposure` and `wind_exposure` properties (I just wanted to remove the setters such that they become read-only).  I originally didn't think about exposing them on `__init__` because they aren't exposed there now but, now that you mention it, that seems like a really good idea.  So people can always create the custom boundary condition and assign it to a Face if they really want to.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        return self._properties
+
+    @property
+    def to(self):
+        """Aperture writer object.
+
+        Use this method to access Writer class to write the aperture in other formats.
+
+        Usage:
+            aperture.to.idf() -> idf string.
```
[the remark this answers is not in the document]
chriswmackey: I have to say that I'm a bit confused why it is set up like this.  Intuitively, I feel that you shouldn't have to pass the object to a method that is being called from that object.  I will put it back the way it is now but maybe we want to do something here similar to what we do with the `properties` and the `host`.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        self._geometry = geometry
+        self._parent = None  # _parent will be set when the Face is added to a Face
+
+        # process the type and boundary condition
+        self.type = type or Window()
+        self.boundary_condition = boundary_condition or Outdoors()
+
+        # initialize properties and writer for extensions
+        self._properties = ApertureProperties(self)
+        self._writer = writer
```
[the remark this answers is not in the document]
chriswmackey: I didn't really touch this part from your original code since it seems that you have a clearer idea about it and it's not something that I really plan on using much with Energy.  The fact that the OpenStudio pathway offers a lot more functionality and that all of these IDF methods will be deprecated by the time our SBIR is finished doesn't make a very compelling case to develop them very robustly or to support them on the plugin level. I will make this change here but I feel that it may be best if you develop out these writers for Radiance first and then I will follow your lead if we want to have a minimal version for energy.

honeybee/aperture.py
```diff
@@ -0,0 +1,506 @@
...
+        """Create an Aperture from vertices with each vertex as an iterable of 3 floats.
+
+        Args:
+            name: Aperture name. Must be < 100 characters.
+            vertices: A flattened list of 3 or more vertices as (x, y, z).
+            type: Aperture type. Default: Window
+            boundary_condition: Boundary condition object (eg. Outdoors, Surface).
+                Default: Outdoors.
+        """
+        geometry = Face3D(tuple(Point3D(*v) for v in vertices))
```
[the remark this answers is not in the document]
chriswmackey: So I re-ran my tests with the most recent version of `ladybug_geometry` and I realized that I must have been mistaken about the reason why the performance was suffering so much.  I [reopened the issue](https://github.com/ladybug-tools/ladybug-geometry/issues/25) and I will implement float casting within the Point2D and Point3D objects now.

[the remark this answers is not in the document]
chriswmackey: This has been addressed in ladybug_geometry now (https://github.com/ladybug-tools/ladybug-geometry/pull/33)

honeybee/aperture.py
chriswmackey: I don't think implementing the IDF equivalents of these objects is nearly as useful as having them in the core for a couple of reasons:
1. It they only exist in honeybee-energy, they become more difficult to coordinate with daylight, radiation, and thermal comfort studies that use Radiance. Here in the core,  both energy and radiance reference the same geometry.
2. Simply having the overhang represented by a depth (as it is in E+) and not having the detailed geometry information for the resulting overhang makes it more difficult to check that the overhang is modeled as intended.  For example, if the window doesn't have a perfectly flat top, you don't get a complete overhang and you want to check/visualize this geometry rather then worrying about how E+ interpreted your inputs.  There's no question about what the geometry is when you have detailed shading geometry.

honeybee/aperture.py
chriswmackey: I disagree that an exception is what you want to happen.  Let's say that you want to generate fins for a facade with 20 rectangular windows and 1 triangular window that sits under gabled roof.  You run this method over the apertures in the Face expecting to get fins for the 20 windows but instead you get an exception because of that pesky triangle in the roof.  You think that maybe you should remove the triangle aperture from the face and add it back afterwards but that is too much work.  So you wrap your whole code in a try/except loop to catch the `ValueError` raised by the one triangle aperture but this means that you don't properly catch all of the other ValueErrors that can happen while creating an overhang (like inputting a non-numerical string for the overhang depth instead of a number). So now you have some bad code that will cause errors.  If you instead read the docstring here and just added a `if overhang is None` in your code, things will be much better and much more resilient.
So, I would argue that, if there's any place to put an exception, it's on the plugin end and I honestly think that is overkill for a Grasshopper plugin because people can visually evaluate what the shade geometry looks like there.  Let them check the result and see if it is what they want rather than stooping them from visualizing the issue in the first place.

honeybee/aperture.py
chriswmackey: Fair point.  I will add this check in

honeybee/aperture.py
chriswmackey: Do you want me to set `raise_exception` to `False` for all ~20 of the checks that I implemented throughout the library?  Or Just these 2?  I feel that we should probably be consistent across the library but I don't have a strong opinion about whether they should all be `False` or all be `True`. I just know that I'm going to forget if these 2 are the only exceptions.

honeybee/aperture.py
chriswmackey: I absolutely agree here.  Now that the `type` and `boundary_condition` are immutable objects, there's no need to duplicate them.  So I removed the `duplicate` methods from them along with the code here.

honeybee/aperturetype.py
chriswmackey: I think this merits some consideration but it might be tough if we ever want to include properties on the types.  For example, it might be nice to store the `fraction_operable` on an `OperableWindow` type in honeybee-energy.  This way, the operable fraction property only gets created for operable windows and doesn't need to exist on the energy properties of every single aperture.  Let's consider this in more detail later.

honeybee/boundarycondition.py
chriswmackey: This is not correct since, after one of our conversations, we agreed that the boundary condition objects are just going to be the names of the objects and not entire objects.  Seeing this example actually makes me think that we made the right decision.  If you think about it, you can see that there would be infinite recursion had the boundary condition object actually been an object that we were sending `to_dict` like what you have here. Having the adjacent face going `to_dict` would cause that adjacent Face's boundary condition to also be called `to_dict`, which would call the original Face's boundary condition `to_dict` and so on.

honeybee/boundarycondition.py
chriswmackey: At first, I thought I might but now I see that it really is not necessary.  I'll remove it.

tests/lockable_test.py
chriswmackey: Good idea. Added

tests/boundary_condition_test.py
chriswmackey: Right!  I just fixed it.

honeybee/typing.py
chriswmackey: I'm not sure why you say that because white spaces are definitely present in the test that I wrote for this method:
https://github.com/ladybug-tools/honeybee-core/pull/10/files#diff-5e17cef320555cc0b9c2ad939932b42dR9
Is a white space not considered an ASCII character in some cases?

honeybee/aperture.py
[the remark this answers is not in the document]
mostaphaRoudsari: This happens for the same reason that you pass the face itself (aka host) to properties. We need several different information from the face. The difference is that the writer will use the object only once and then it will be garbage collected so storing it as an object is not really useful.

honeybee/aperture.py
mostaphaRoudsari: If it wasn't clear I'm in favor of having them in the core library but this comment wasn't about that. What I meant is to add a property like `shading` to `aperture` to collect shading objects so you can do something like `aperture.shading` and it returns all the fins, overhangs and the rest of the shading objects. Right now you have to collect these objects in a separate list somewhere.

honeybee/aperture.py
mostaphaRoudsari: I agree that this doesn't deserve an exception and I suggested a warning (warning.warns('No overhang is created.')). This will not raise an exception.

honeybee/aperture.py
mostaphaRoudsari: I mean it for all. Basically I think when one uses `face.is_planar()` it should not throw an exception by default. It should return True or False.

honeybee/aperturetype.py
mostaphaRoudsari: I agree. Let's keep this for later.

honeybee/boundarycondition.py
mostaphaRoudsari: You are right. Let's see how this implementation will lead us. I know the logic that lead us to this decision but it still doesn't _feel_ right.

honeybee/typing.py
mostaphaRoudsari: Hmm. You are right. I should have run the wrong code. I should say that I still like the regular expression version. It is more explicit than ord(i) < 128. But it does work so 👍.

honeybee/aperture.py
chriswmackey: Sorry that I misunderstood.  I was considering before this but it adds another level of complexity that I am not sure is helpful because people can already assign shades to zones.  If we allow people to assign shades to apertures, it also seems weird to not let them assign shades to faces, which means that we now have to keep track of shades assigned to 4 objects (Aperture, Face, Room, Model) instead of just 2 (Room and Model).

Also, I think the Room level is really where we want people to assign them since aperture-level shading really doesn't have a meaning for E+.  Plus I am sure that certain overhangs are big enough that you really want to account for the fact that they shade neighboring windows in the room or have an effect beyond just that window.
If you want something like this that auto-assigns shades to an object, we can add some methods on the Room to generate shades across all of the windows of a certain orientation or something like that.

honeybee/aperture.py
chriswmackey: Ah. I should have read more carefully.  Ok, I can import warnings and add a warning like what you say.

honeybee/aperture.py
chriswmackey: Ok. Good to know.  I remember you saying a while back that we should use the practice of always setting `raise_exception` to `True` for functions that check things but I see your opinion has changed here.  I will use a practice of setting it to `False` here in `honeybee_core` and I'll consider using the same practice elsewhere across the libraries. Or, if you think defaulting to `False` is not good practice for other libraries, let me know if there are any reasons why `honeybee-core` is the exception to the rule.

mostaphaRoudsari: I have probably mis-communicated what I meant. These are two different cases in my mind.

Case 1 is a check for making a follow up decision and case 2 is the check as a requirement. `is_planar` to me is more of a case 1. If it was `check_planarity` then I would have expected it to raise an exception is it is not planar.

honeybee/aperture.py
mostaphaRoudsari: This logic works well for EnergyPlus but it is not how shadings are designed most of the times and also is not how it will be used in a daylight model. I think we have to at least keep track of them in 3 objects. We can remove the assignment for Face because it will be the same as assigning them to zones in both energy and daylight models.

Here is how I would categorize them.

1. One can add shadings to Aperture (which is the most common case for blinds, fins, etc) and what will be used for dynamic daylight modeling. This will be added as a Zone shading in energy models.
2. If not a specific aperture but a specific Zone then it can be assigned to the Zone. This will be used as exterior shades in daylight and is the ideal case for energy models.
3. Model level shading will be only for context shading. Will be treated the same in both daylight and energy models.

Not to make it more complicated but there will be a need for dynamic shading objects which will be defined differently in energy versus daylight models. That can probably be handled by `honeybee-energy` and `honeybee-daylight`.
 
If you think it will be too much to ask for this PR (which I would agree) we can open an issue to discuss this further in the future.

chriswmackey: Ok, I'm just glad to hear you agree that it should not be a part of this PR since I don't think that I'm really be best one to implement it.  You have a much clearer idea of why you need this functionality for Radiance than I do and, as I said previously, we really don't need/want it for energy or thermal comfort. So please feel free to open an issue if you want to tackle this later and here's a list of things that should be addressed when you add support for aperture-specific shading:

1. Make it clear to the user that aperture-level shading only applies to a single aperture in Radiance (and not the apertures that are right next to it) but it applies to all apertures of the Room for energy/thermal comfort.
2. Give some warnings in the docstrings that Aperture-level overhangs and fins can be dangerous when you have two apertures that are really close to one another.  For example, a fin that looks like it is in between two apertures that are very close to one another is actually only shading one of the two apertures.
3. Have clear explanations for why the API is not consistent in how it deals with shades (ie. they can be applied to Models, Rooms, and Apertures but not Faces).  And, even though you can still generate shades from Face geometry, you cannot assign it to the Face.
4. Handle the fact that it is not clear whether Aperture-level shading is on the inside or the outside of a Room.  For example, you might generate shades for an aperture without a parent and the shades look like they are on a certain side of the aperture.  Then, the aperture gets added to a Room and the aperture face gets flipped to ensure that it is facing outwards from the Room.  This reverses the inside/outside relationship of the shades assigned to the aperture.

mostaphaRoudsari: I opened an issue with link to these comments and leave the conversation for the future.

https://github.com/ladybug-tools/honeybee-core/issues/11

honeybee/aperture.py
chriswmackey: I'm sorry to say that I am even more confused now.  So the decision about how the method is named should determine whether `raise_exception` is `True` or `False`?  So maybe we should be changing the name of the method to 'check_planarity` rather than the `raise_exception` default?
I think that I'll just leave it the way it is now since I don't think that you have gotten the chance to look at the other ~20 checks methods in this PR.  After you have taken a look at those, then let me know what you think here.

honeybee/_lockable.py
```diff
@@ -0,0 +1,81 @@
...
+        cls.lock = lock
+    if not hasattr(cls, 'unlock'):  # allow developers to add their own unlock method
```
chriswmackey: Add a exception when cls has __slots__ and there's not _locked variable assigned.

honeybee/face.py
```diff
@@ -1,158 +1,875 @@
...
+        self._boundary_condition = Surface(other_face)
+        other_face._boundary_condition = Surface(self)
+        # set the apertures to be adjacent to one another
+        if len(self._apertures) > 0:
+            found_adjacencies = []
+            for aper_1 in self._apertures:
+                for aper_2 in other_face._apertures:
+                    if aper_1.center.distance_to_point(aper_2.center) <= tolerance:
+                        aper_1.set_adjacency(aper_2)
+                        found_adjacencies.append(True)
```
mostaphaRoudsari: why don't you just use a counter here instead of creating a list? As far as I can see the only use is to check the length of found adjacencies against the apertures.

honeybee/aperture.py
chriswmackey: I added the warnings in but I think that I underestimated how annoying they would be while running all of the tests.  You are going to have to show me how to turn the warnings off at some point. I can easily see myself getting to the point where I can't see the real errors amidst all of the warnings, especially if we are going to use them for cases like this one where I still a bit unsure that the warning is actually helpful.

honeybee/face.py
```diff
@@ -1,158 +1,875 @@
...
+                to scale. If None, it will be scaled from the World origin (0, 0, 0).
+        """
+        self._geometry = self.geometry.scale(factor, origin)
+        for ap in self._apertures:
+            ap.scale(factor, origin)
+        for dr in self._doors:
+            dr.scale(factor, origin)
+        self._punched_geometry = None  # reset so that it can be re-computed
+
+    def are_sub_faces_valid(self, tolerance, angle_tolerance, raise_exception=True):
```
mostaphaRoudsari: we can check the name of methods with `are_*` to `validate_*` or `check_*`

chriswmackey:
@mostaphaRoudsari ,
All of what we talked about has been addressed and this one is ready for continued review.

honeybee/shade.py
```diff
@@ -0,0 +1,284 @@
...
+    def to(self):
+        """Shade writer object.
+
+        Use this method to access Writer class to write the shade in different formats.
+
+        Usage:
+            shade.to.idf(shade) -> idf string.
+            shade.to.radiance(shade) -> Radiance string.
+        """
+        return writer
```
mostaphaRoudsari: This should probably be set to `NotImplementedError()` until we implement the writers.

honeybee/aperture.py
mostaphaRoudsari: Would this work: https://stackoverflow.com/a/14463321/4394669

honeybee/room.py
```diff
@@ -0,0 +1,605 @@
...
+            faces.append(Face('{}_Face{}'.format(name, i), face,
+                              get_type_from_normal(face.normal, roof_angle, floor_angle),
+                              get_bc_from_position(face.center, ground_depth)))
+        room = cls(name, faces)
+        room._geometry = polyface
+        return room
+
+    @classmethod
+    def from_box(cls, name, width=3.0, depth=6.0, height=3.2,
+                 orientation_angle=0, origin=Point3D(0, 0, 0)):
```
mostaphaRoudsari: I know this is a left-over from honeybee room but `rotation_angle` might be a better choice than `orientation_angle`.

honeybee/room.py
```diff
@@ -0,0 +1,605 @@
...
+
+        Args:
+            shade: A shade face to add to the outdoors of this room.
+        """
+        assert isinstance(shade, Shade), \
+            'Expected Shade for outdoor_shade. Got {}.'.format(type(shade))
+        shade._parent = self
+        self._outdoor_shades.append(shade)
+
+    def generate_mesh_grid(self, x_dim, y_dim=None, offset=1.0):
```
mostaphaRoudsari: I know the results will be a mesh but not sure if we need to include it in the name of the method. You can just call it `generate_grid`.

honeybee/model.py
```diff
@@ -0,0 +1,480 @@
...
+        apertures
+        doors
+        has_ngon_apertures
+        had_ngon_doors
+    """
+    __slots__ = ('_name', '_name_original', '_rooms', '_faces', '_shades',
+                 '_apertures', '_doors', '_room_names', '_face_names', '_shade_names',
+                 '_aperture_names', '_door_names', '_properties')
+
+    def __init__(self, name, objects):
```
mostaphaRoudsari: @chriswmackey, We may want to discuss model over a call. There are a number of design decisions here that I want to know more why you made those choices.

1. Not sure if we need `x` and `x_names` properties. I would say a `x` method with a `names_only` argument looks like a cleaner choice to me.
2. Not a big fan of collecting objects as dictionary. I assume this decision was made to ease the process of returning a specific room, face or aperture but this can have some major performance downsides for us specially for large models. See this presentation for a reference: https://youtu.be/7eeEf_rAJds and since your next question is how is that possible when Python instances create \_\_dict__ unless using \_\_slots__? Well I don't know the exact answer but someone asks the same question at 33:40. 
2.1. if my assumption about access is correct using a quick search through a list or tuple just works as well. Specially if we have the separation based on type done in the model.

honeybee/model.py
```diff
@@ -0,0 +1,480 @@
...
+        self._apertures = {}
+        self._doors = {}
+        self._room_names = []
+        self._face_names = []
+        self._shade_names = []
+        self._aperture_names = []
+        self._door_names = []
+        for obj in objects:
+            if isinstance(obj, Room):
+                self.append_room(obj)
```
mostaphaRoudsari: `append_*` is totally fine but to be consistent with other objects we might use `add_*` instead.

honeybee/model.py
```diff
@@ -0,0 +1,480 @@
...
+    __slots__ = ('_name', '_name_original', '_rooms', '_faces', '_shades',
+                 '_apertures', '_doors', '_room_names', '_face_names', '_shade_names',
+                 '_aperture_names', '_door_names', '_properties')
+
+    def __init__(self, name, objects):
+        """A collection of Rooms, Faces, Apertures, and Doors for an entire model.
+
+        Args:
+            name: Model name. Must be < 100 characters.
+            objects: A list of honeybee Rooms, Faces, Apertures, and Doors.
```
mostaphaRoudsari: should model has an input for north?

honeybee/model.py
```diff
@@ -0,0 +1,480 @@
...
+        """Append another Model object to this model."""
+        assert isinstance(other_model, Model), \
+            'Expected Model. Got {}.'.format(type(other_model))
+        for room in other_model.rooms:
+            self.append_room(room)
+        for face in other_model.faces:
+            if not face.has_parent:
+                self.append_face(face)
+        for shade in other_model.shades:
+            if not shade.has_parent:
```
mostaphaRoudsari: Shouldn't this always be the case as this shade is part of the model? Same comment about face and aperture.

honeybee/model.py
```diff
@@ -0,0 +1,480 @@
...
+    def to(self):
+        """Model writer object.
+
+        Use this method to access Writer class to write the model in other formats.
+
+        Usage:
+            model.to.idf(model) -> idf string.
+            model.to.radiance(model) -> Radiance string.
+        """
+        return writer
```
mostaphaRoudsari:
```suggestion
        raise NotImplementedError()
```

honeybee/model.py
mostaphaRoudsari: I see that you are collecting all shades, faces and apertures so now it makes sense but I'm not sure if we need to do that. If we have implemented access to faces effectively in room objects then this is the time to use them in my opinion. What I'm trying to say is that we can only collect orphaned faces under `_faces` and then for `faces` method we can collect all the faces from rooms and return them with orphaned faces. Something like this: `itertools.chain((room.faces for room in self.rooms), self._faces)`

honeybee/boundarycondition.py
```diff
@@ -1,125 +1,163 @@
...
 
     def __repr__(self):
         return self.name
 
 
 class Surface(_BoundaryCondition):
+    """Bondary condition when an object is adjacent to another object."""
 
-    __slots__ = ()
+    __slots__ = ('_boundary_condition_objects')
```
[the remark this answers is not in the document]
chriswmackey: Good catch.

honeybee/door.py
```diff
@@ -0,0 +1,334 @@
...
+    def from_vertices(cls, name, vertices, boundary_condition=None):
+        """Create a Door from vertices with each vertex as an iterable of 3 floats.
+
+        Args:
+            name: Door name. Must be < 100 characters.
+            vertices: A flattened list of 3 or more vertices as (x, y, z).
+            boundary_condition: Boundary condition object (eg. Outdoors, Surface).
+                Default: Outdoors.
+        """
+        geometry = Face3D(tuple(Point3D(*v) for v in vertices))
```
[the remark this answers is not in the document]
chriswmackey: Haha! Yes, but no more than 4 vertices. So don't be _that_ impressed :)

honeybee/door.py
```diff
@@ -0,0 +1,334 @@
...
+
+        Args:
+            other_aperture: Another Aperture object to be set adjacent to this one.
+        """
+        assert isinstance(other_door, Door), \
+            'Expected Door. Got {}.'.format(type(other_door))
+        self.boundary_condition = Surface(other_door)
+        other_door.boundary_condition = Surface(self)
+
+    def move(self, moving_vec):
```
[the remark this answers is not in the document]
chriswmackey: I can see 4 arguments for having this method:
1. There are likely many times where you generate a door/aperture in the center of a face but you want to move it to the side of the face.
2. Having the ability to move a door may help us manage doors that share an edge with their parent (something that is legal in E+ but could be problematic in geometry interfaces like Rhino).
3. Remember that the parent face still has an `are_sub_faces_valid` method that can catch if the door has been moved to an illegal location.
4. We probably at least need a hidden version of this method, which will be called by the parent face when it is moved.

honeybee/door.py
```diff
@@ -0,0 +1,334 @@
...
+            origin: A ladybug_geometry Point3D for the origin around which the
+                object will be rotated.
+        """
+        self._geometry = self.geometry.rotate(axis, math.radians(angle), origin)
+
+    def rotate_xy(self, angle, origin):
+        """Rotate this Door counterclockwise in the world XY plane by a certain angle.
+
+        Args:
+            angle: An angle in degrees.
```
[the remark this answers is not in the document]
chriswmackey: I think the important thing is that each library is consistent within itself.  Except for `ladybug-geometry`, all other Ladybug Tools libraries currently use degrees as the default angle of measurement.  This is true for altitude/azimuth angles in ladybug-core and it is true of all angle measurements currently in this honyebee-core PR.  The main reasons why ladybug-geometry is the black sheep in the Ladybug Tools family is that using degrees there introduces several extra inefficiencies both because it involves an extra step for all of the internal trigonometric functions (for which there are a lot in ladybug-geometry) and because the geometry formats used by CAD interfaces are typically using radians. If you are willing to change over ladybug-core to use radians instead of degrees, we can consider a global standard for all of the libraries but I am personally in favor of keeping ladybug-geometry as the black sheep and having all other libraries use degrees.

honeybee/door.py
```diff
@@ -0,0 +1,334 @@
...
+            return False
+        return True
+
+    @property
+    def properties(self):
+        """Door properties, including Radiance, Energy and other properties."""
+        return self._properties
+
+    @property
+    def to(self):
```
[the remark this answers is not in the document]
chriswmackey: Done!

honeybee/properties.py
```diff
@@ -79,4 +49,268 @@ def ToString(self):
...
+    Door properties. This class will be extended by plugins.
+
+    Usage:
+        door = Door('Front Door', geometry)
+        door.properties -> DoorProperties
+        door.properties.radiance -> DoorRadianceProperties
+        door.properties.energy -> DoorEnergyProperties
+    """
+
+    def __init__(self, host):
```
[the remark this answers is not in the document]
chriswmackey: I was considering this but I was also realizing that the `__init__` is really only 2 lines of Python and the majority of the code there is just docstrings.  It also seemed helpful here for the documentation to explain that the host is a `Face that hosts these properties` or a `Model that hosts these properties` since I imagine someone getting confused the first time that they look at how properties work in the API and see that it's not like any other object. I can move it to the base class if you think that I'm overthinking it but I just wanted to explain my reasoning.

honeybee/face.py
```diff
@@ -1,158 +1,875 @@
...
+        self._punched_geometry = None
         self._apertures = []
+        self._doors = []
+
+        # set face type based on normal if not provided
+        self.type = type or get_type_from_normal(geometry.normal)
+
+        # set boundary condition by the relation to ground plane if not provided
+        self.boundary_condition = boundary_condition or \
+            get_bc_from_position(geometry.center)
```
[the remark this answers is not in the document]
chriswmackey: Ok, I can change it to only assign the ground bc if all positions are at or below the ground depth.  I think the docstring changed from when you wrote this comment since it doesn't say that you can change the ground depth any more.  And this makes sense  given that, on the face-level, you can just input a Ground bc on `__init__`.  However, when you make a Room from a polyface, there's an input argument to change the ground plane.

honeybee/face.py
```diff
@@ -1,158 +1,875 @@
...
+    @property
+    def has_parent(self):
+        """Boolean noting whether this Face has a parent Room."""
+        return self._parent is not None
 
-        args:
-            name: Face name.
-            vertices: A flattened list of 3 or more vertices as (x, y, z).
-            face_type: Face type (e.g. wall, floor).
+    def horizontal_orientation(self, north_vector=Vector2D(0, 1)):
```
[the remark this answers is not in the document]
chriswmackey: I think it's possible that you misunderstood what this method is doing.  It isn't setting the north at all. It's just allowing you to figure out the orientation of a face given a north vector.   So you can set the north on the model and then you can use this method to figure out the orientation of an individual face by inputting the model north vector to it. I just think you want to have something to help you on the face-level, especially if you haven't yet added all your faces to a model object and you want to write a script that does something like" set all of the faces of _X_ orientation to have _Y_ glazing ratio."
On a related note, do you think the model north vector should be something that is set on the model object in honeybee-core?  Or should the north live within the model radiance or energy properties?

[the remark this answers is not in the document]
mostaphaRoudsari: Ah I see. Makes total sense and will be very useful.

I think it should be in the core library as the north orientation is useful for both energy and daylight models.

honeybee/door.py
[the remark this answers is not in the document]
mostaphaRoudsari: Agree that 1 and 2 are useful cases. So maybe to ensure that the door stays in the same plane then we probably should ask for a 2d vector to move the subface geometry inside the parent face plane.

honeybee/door.py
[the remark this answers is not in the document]
mostaphaRoudsari: Sounds good to me. Where do you think we should document this?

honeybee/properties.py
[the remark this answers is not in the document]
mostaphaRoudsari: The reasoning behind the docstring is valid but I would say in case we ever wanted to make a change to \_\_init__ method for these classes reusing the \_\_init__ is the right approach which will save us from editing the code multiple times. I think a generic `Geometry / object that hosts these properties` will do the job for the API.

honeybee/face.py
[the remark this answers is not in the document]
mostaphaRoudsari: >  I think the docstring changed from when you wrote this comment since it doesn't say that you can change the ground depth any more.

See line 79. I meant the comment in the code not the docstring.

This has been one of those issues that comes up once in a while when someone is modeling a geometry on a slope, etc. I think this is an issue that we can revisit later.

honeybee/face.py
```diff
@@ -1,158 +1,875 @@
...
+        are_sub_faces_valid() method can be used afterwards.
+
+        Args:
+            aperture: An Aperture to add to this face.
+        """
+        assert isinstance(aperture, Aperture), \
+            'Expected Aperture. Got {}.'.format(type(aperture))
+        self._acceptable_sub_face_check(Aperture)
+        aperture._parent = self
+        if self.normal.angle(aperture.normal) > math.pi / 2:  # reversed normal
```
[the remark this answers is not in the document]
chriswmackey: Yes, that is true and we can add an angle_tolerance input but I know that it will get annoying for people working with the Python API to have to specify an angle tolerance every time they want to add an aperture.  If this were a case where we could use a default tolerance of 0  and everything would usually be ok, I would do that but most faces that aren't perfectly vertical or horizontal will raise an exception in this case simply because of Python floating point tolerance.  I'm also not sure that adding an angle_tolerance check here will really add much more efficiency to the checks since you probably want to run the `check_are_sub_faces_valid()` anyway in order to verify that the sub-faces lie within the boundary of the parent face.

honeybee/face.py
chriswmackey: Thanks for catching this. I had the sense while writing this that it was not the correct way to do it but I had forgotten what the right way is.  I'll replace it with a counter now.

honeybee/door.py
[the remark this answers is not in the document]
chriswmackey: Glad that you agree with points 1 and 2. While I know that it sounds good in an abstract sense to use a Vector2D, I can guarantee you that it's not nearly as intuitive as you think and, as a result, it is more prone to errors. This is primarily because there's no guarantee that the plane of a given Face has an X or Y-axis pointing in a certain direction.  For example, when you rotate a Face, you rotate the plane with it and, while this is great for cases like a shoebox model and ensuring that the generated mesh grid is always oriented along the dimensions of the box, it also means that you easily end up with cases where the Face plane has a Y-axis pointing in the negative Z direction, which is the opposite of what you would like for a case like moving a door. I use the rotation as an example but there are many other ways that the axis can change from what you expect and it's a very common condition when geometry is coming from CAD interfaces and we have to generate the plane from the vertices.  At least in 3D, you know what direction is up and which is down, which makes it much more likely that you use the right vector. If you still eel strongly about it, I can add a method that accepts a Vector2D but I am going to insist that the 3D version be kept and I would write all of the warnings that I noted above into the method that uses the Vector2D.

honeybee/door.py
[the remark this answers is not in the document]
chriswmackey: Good question.  Maybe we should just write it into the [global contributing Readme](https://github.com/ladybug-tools/contributing/blob/master/README.md) that all Ladybug Tools libraries are expected to use degrees except `ladybug-geometry`. Let me know if this seems like the right place and I can add it.

honeybee/properties.py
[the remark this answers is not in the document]
chriswmackey: Ok.  I'll move it to the base class then.

honeybee/room.py
```diff
@@ -0,0 +1,605 @@
...
+            self._faces.append(face)
+        self._geometry = None  # will be calculated from faces or added by classmethods
+        self._indoor_shades = []  # shades can be assigned later
+        self._outdoor_shades = []  # shades can be assigned later
+
+        # initialize properties for extensions
+        self._properties = RoomProperties(self)
+
+    @classmethod
+    def from_faces(cls, name, faces, tolerance, angle_tolerance):
```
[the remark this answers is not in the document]
chriswmackey: The main reason for this not a performance issue.  Primarily, it's that, if we force this to be the method that everyone has to use, it won't be be impossible to create a Room object if you don't know the tolerance under which the geometry was originally created.  This makes Rooms the only geometry object that is tied to a tolerance and it means that we loose the opportunity for the honeybee-core library to accept geometry from sources where the original tolerance has been lost.  If we are certain that we only want honeybee-core to accept geometry from certain sources, we can go this route but I have been hoping that honeybee-core is destined for more inter-operability than that.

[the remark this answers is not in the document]
chriswmackey: Thinking about it more, I could move the two tolerance arguments into init and just set them equal to None by default, which will effectively do the same thing that the init method does now. And, if someone inputs a numerical tolerance, it will run the solidity checks like the current from_faces method. I had originally thought that having two separate methods to initialize the Room was more explicit but maybe you are right that having users recognize the tolerance as None is better. Let me know your thoughts.

honeybee/room.py
```diff
@@ -0,0 +1,605 @@
...
+        for face in faces:
+            assert isinstance(face, Face), \
+                'Expected honeybee Face. Got {}'.format(type(face))
+        face_geometries = tuple(face.geometry for face in faces)
+
+        # try to get a closed volume between the faces
+        if tolerance == 0:
+            room_polyface = Polyface3D.from_faces(face_geometries)
+        else:
+            room_polyface = Polyface3D.from_faces_tolerance(face_geometries, tolerance)
```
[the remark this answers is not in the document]
chriswmackey: After reading this, I am guessing that you will be in favor of the solution that I suggested above for honeybee-core but I will wait for you to confirm.
The main difference between those two classmethods is how they determine a shared vertex between the set of faces.  The first one (`from_faces`) uses the Python `index()` method to find a matching point, meaning that things need to match within 0 tolerance and that a built-in Python method, which is optimized for the task is being used.  For the second one (`from_faces_tolerance`), the method loops through all of the other vertices to test each one if the difference x, y, and z values are all less than or equal to the tolerance.  This is not as optimized but it allows you to sense whether a polyface is closed to within a non-zero tolerance.  I thought that I was being more explicit by having separate classmethods for these two cases but maybe this is a case where the next line of Python Zen applies:
> Simple is better than complex.
Perhaps the `index()` method really isn't that much more optimized than the tolerance test that I mentioned and I should just only have the one method, which can work with a 0 tolerance input anyway.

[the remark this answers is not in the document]
chriswmackey: I found one more bit of Python Zen to support your point:
> There should be one-- and preferably only one --obvious way to do it.

If you confirm that these points apply to our case here, I will implement both of them

honeybee/room.py
chriswmackey: Actually, `rotation_angle` was what you originally had for the older honeybee Room but I thought it didn't provide any clarity as to whether the angle was like a standard rotation operation, which typically follows right-hand rule logic to go counterclockwise, or whether it was trying to match a specific cardinal orientation, which typically starts from the north and goes clockwise.  I looked at your code ans it was the latter so I thought `orientation_angle` was clearer.  If there was some logic to your original choice of the word `rotation_angle` that was lost on me, please elaborate.

honeybee/room.py
chriswmackey: I am fine with that if you think `generate_grid` is clear enough.  I will make the change and I'll make sure the docstring elaborates that the output is a mesh.

honeybee/model.py
chriswmackey: @mostaphaRoudsari ,
It sounds like a call is in order. I really put the model class together relatively quickly, almost more as a record for all of the possible checks that we would like to build into it rather than anything that was optimized for accessing or storing other objects.

FYI, the reason for the `x_names` variables was to sense when there was a duplicate-named objects in the model since the dictionary key would be overwritten when this is the case.

I will take a look at that youtube video and I recognize that, if you are being mindful of the performance in this case after you have previously pushed to run a lot of geometry checks implicitly, then this dictionary situation must be a pretty big performance issue 🙂

honeybee/model.py
chriswmackey: NotImplemented error has been implemented!

honeybee/model.py
chriswmackey: Ah, I got my answer to my comment above.  I'll move the one that I put on honeybee-energy to the honeybee-core one instead.

honeybee/face.py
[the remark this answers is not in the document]
chriswmackey: I made the comment more explicit that it is a zero ground plane by default. And I agree that checking all of the vertices is better practice.

honeybee/model.py
chriswmackey: Let's have a longer call to discuss this after we have decided on the exact way that we want to organize the objects (if not using dictionaries).

honeybee/model.py
chriswmackey: Thanks again for the reference video.  What the presenter said in regards to the question about dicted classes makes sense and I realize that there is a dark world of how the code is compiled that relies heavily on correctly interpreting people's 

I just wanted to outline the alternatives to the dictionary that seem worth considering here:

- two lists for each geometry object type - One with the names of the objects and another with the instances of the objects.  This might make lookups of objects by name a little slower since you will need to first find the index in the name list and then use that index to return the instance in the other list.  However, it would definitely make the size of the Model object in memory smaller since the two lists are pretty certainly less than the dictionary.  It also avoids the need for a separate lists of names from the one used to lookup objects, allowing use to check for duplicate names by just using the list with names.  Also, by using lists, it is relatively straightforward to add new objects to the model (we just append them to the two lists) and this also makes it easier to add all of the elements of parent objects to their respective lists (like all of the faces of a room to the "faces" list).  If optimized fast lookups is what we are willing to compromise on, this seems like the best solution.  And, thinking about it more, I am not sure that we will really use the lookups much beyond the check for adjacent objects with Surface boundary conditions.  Otherwise, people can easily access the faces of a room from the room object and there's no need to lookup the face on the model.

- namedtuple for each geometry object type - According to the talk, this does not have the same performance/memory issues as a dictionary and it will still allow for quick lookups.  However, I think the biggest drawback of this approach is that it's a tuple under the hood and not a list.  As a result, it becomes really intense to add new objects to the model or add all of the elements of a parent object since the tuple under the hood is immutable.  Personally, I don't think that having such optimized lookups is worth it here.

- relying on parent objects to store the only reference to their children - This will make lookups by name extremely painful.  However, it is the smallest size in memory.

honeybee/model.py
chriswmackey: See my comment above about "Relying on parent objects to store the only reference to their children" for my thoughts on your proposal here.  If we think that we are planning on not using any of the checking methods, then what you propose here is worth considering.  But otherwise, the check for duplicate names and adjacent objects will have so much overhead.

honeybee/door.py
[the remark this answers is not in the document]
mostaphaRoudsari: Sounds good. I still think we should have it somewhere else for both users and contributors but adding it to contributing README should be a good start. Maybe we should add s note about units for each library in its README. It might get complicated now so I'm not sure.

honeybee/room.py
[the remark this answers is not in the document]
chriswmackey: You confirmed over our video call so in it goes!

honeybee/door.py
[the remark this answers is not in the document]
chriswmackey: Ok, I added a whole section to **Contributing.MD** on [unit conventions](https://github.com/ladybug-tools/contributing/blob/master/README.md#unit-conventions), including the fact that we assume SI units by default.  "Complicated" is the word that I would choose if we wanted to have different contributing standards listed on each repo.  Let's try to have one (and preferably only one) set of contributing guidelines and only resort to a different set of standards on each repo if we have to.

honeybee/room.py
mostaphaRoudsari: To me both terms are very similar. Don't have a strong opinion. If this is closer to what the angle does then let's keep it as is.

honeybee/model.py
mostaphaRoudsari: let's talk about all the comments about the model over the call. I actually think the third option is the best and don't see the look up to be as painful. That's how almost all d3js look-ups work for example.

honeybee/face.py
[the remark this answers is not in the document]
mostaphaRoudsari: That was not what I meant but I see that what you are trying to do here is different then ensuring the faces are on the same plane so it should be good.

I think on the model, room and face we should have `validate` methods that does all the checks. It doesn't need to run by default but it is an option for to ensure model quality.

honeybee/door.py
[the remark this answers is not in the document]
mostaphaRoudsari: I see your points but it should not be as complicated. Since you have methods that sorts the points, etc it should not be hard to align the plane with bounding box of the face if needed but for now let's keep it as is and see if anyone ever complains about it.

honeybee/face.py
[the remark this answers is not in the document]
chriswmackey: If you don't mind, I'm just going to keep this unresolved since I agree that we can expose methods on room and model that check all children faces.  It's a simple thing to add given that the Face, Aperture, and Door, already have the check methods so I can do it in this PR.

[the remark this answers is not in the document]
chriswmackey: Done!  I added in this capability on both the Room and the Model.

honeybee/model.py
chriswmackey: Ok, we can discuss once I finish the tests and merge it in.  And, just so that what we are merging something that we are actually considering, I switched over the model object to use the two lists (option 1) instead of the previous dictionary. On the call, we can go through option 3 and see how it affects each of the check methods.  As, I said before, I think option 3 is fine if we rarely see ourselves running the check methods.  I imagine that D3 might not have the same need to check things like the presence of adjacent objects in the model, duplicate names, non-adjacent air walls, and the other ~15 checks that are currently built into the Model object.

PR #13, opened 2019-07-26
feat(model): Adding to_dict and from_dict to Model

chriswmackey:
This commit also includes edits to the other geometry objects in order to be able to support the use of to_dict methods to generate the Model schema.

@mostaphaRoudsari ,
The code edits here are ready to be reviewed, though I'll still be adding more tests for the Model object over the next day.

### DIFF
--- a/honeybee/aperture.py
+++ b/honeybee/aperture.py
@@ -1,25 +1,30 @@
 # coding: utf-8
 """Honeybee Aperture."""
+from ._basewithshade import _BaseWithShade
 from .properties import ApertureProperties
 from .aperturetype import aperture_types
 from .boundarycondition import boundary_conditions, Outdoors, Surface
 from .shade import Shade
-from .typing import valid_string
 import honeybee.writer as writer
 
 from ladybug_geometry.geometry3d.pointvector import Point3D, Vector3D
-from ladybug_geometry.geometry3d.plane import Plane
 from ladybug_geometry.geometry3d.face import Face3D
 
 import math
 
 
-class Aperture(object):
-    """A single planar aperture face.
+class Aperture(_BaseWithShade):
+    """A single planar Aperture in a Face.
 
     Properties:
         name
-        name_original
+        display_name
+        type
+        boundary_condition
+        indoor_shades
+        outdoor_shades
+        parent
+        has_parent
         geometry
         vertices
         upper_left_vertices
@@ -28,14 +33,9 @@ class Aperture(object):
         center
         area
         perimeter
-        type
-        boundary_condition
-        parent
-        has_parent
     """
     TYPES = aperture_types
-    __slots__ = ('_name', '_name_original', '_geometry', '_parent',
-                 '_type', '_boundary_condition', '_properties')
+    __slots__ = ('_geometry', '_parent', '_type', '_boundary_condition')
 
     def __init__(self, name, geometry, type=None, boundary_condition=None):
         """A single planar aperture in a face.
@@ -47,9 +47,7 @@ class Aperture(object):
             boundary_condition: Boundary condition object (Outdoors, Surface).
                 Default: Outdoors.
         """
-        # process the name
-        self._name = valid_string(name, 'honeybee aperture name')
-        self._name_original = name
+        _BaseWithShade.__init__(self, name)  # process the name
 
         # process the geometry
         assert isinstance(geometry, Face3D), \
@@ -63,6 +61,33 @@ class Aperture(object):
 
         # initialize properties for extensions
         self._properties = ApertureProperties(self)
+
+    @classmethod
+    def from_dict(cls, data):
+        """Initialize an Aperture from a dictionary.
+
+        Args:
+            data: A dictionary representation of an Aperture object.
+        """
+        # check the type of dictionary
+        assert data['type'] == 'Aperture', 'Expected Aperture dictionary. ' \
+            'Got {}.'.format(data['type'])
+
+        ap_type = aperture_types.by_name(data['aperture_type'])
+        if data['boundary_condition']['type'] == 'Outdoors':
+            boundary_condition = Outdoors.from_dict(data['boundary_condition'])
+        elif data['boundary_condition']['type'] == 'Surface':
+            boundary_condition = Surface.from_dict(data['boundary_condition'], True)
+        else:
+            raise ValueError(
+                'Boundary condition "{}" is not supported for Apertures.'.format(
+                    data['boundary_condition']['type']))
+        aperture = cls(data['name'], Face3D.from_dict(data['geometry']),
+                       ap_type, boundary_condition)
+        if 'display_name' in data and data['display_name'] is not None:
+            aperture._display_name = data['display_name']
+        aperture._recover_shades_from_dict(data)
+        return aperture
 
     @classmethod
     def from_vertices(cls, name, vertices, type=None, boundary_condition=None):
@@ -79,20 +104,40 @@ class Aperture(object):
         return cls(name, geometry, type, boundary_condition)
 
     @property
-    def name(self):
-        """The aperture name (including only legal characters)."""
-        return self._name
+    def type(self):
+        """Object for Type of Aperture (ie. Window)."""
+        return self._type
+
+    @type.setter
+    def type(self, value):
+        assert value in self.TYPES, '{} is not a valid aperture type.'.format(value)
+        self._type = value
 
     @property
-    def name_original(self):
-        """Original input name by user.
+    def boundary_condition(self):
+        """Boundary condition of this aperture."""
+        return self._boundary_condition
 
-        If there are no illegal characters in name then name and name_original will
-        be the same. Legal characters are ., A-Z, a-z, 0-9, _ and -.
-        Invalid characters are automatically removed from the original name for
-        compatability with simulation engines.
-        """
-        return self._name_original
+    @boundary_condition.setter
+    def boundary_condition(self, value):
+        if not isinstance(value, Outdoors):
+            if isinstance(value, Surface):
+                assert len(value.boundary_condition_objects) == 3, 'Surface boundary ' \
+                    'condition for Aperture must have 3 boundary_condition_objects.'
+            else:
+                raise ValueError('Aperture only supports Outdoor or Surface boundary '
+                                 'condition. Got {}'.format(type(value)))
+        self._boundary_condition = value
+
+    @property
+    def parent(self):
+        """Parent Face if assigned. None if not assigned."""
+        return self._parent
+
+    @property
+    def has_parent(self):
+        """Boolean noting whether this Aperture has a parent Face."""
+        return self._parent is not None
 
     @property
     def geometry(self):
@@ -147,38 +192,6 @@ class Aperture(object):
         """The perimeter of the aperture."""
         return self._geometry.perimeter
 
-    @property
-    def type(self):
-        """Object for Type of Aperture (ie. Window)."""
-        return self._type
-
-    @type.setter
-    def type(self, value):
-        assert value in self.TYPES, '{} is not a valid face type.'.format(value)
-        self._type = value
-
-    @property
-    def boundary_condition(self):
-        """Boundary condition of this aperture."""
-        return self._boundary_condition
-
-    @boundary_condition.setter
-    def boundary_condition(self, value):
-        assert isinstance(value, (Outdoors, Surface)), \
-            'Aperture only supports Outdoor or Surface boundary condition. ' \
-            'Got {}'.format(type(value))
-        self._boundary_condition = value
-
-    @property
-    def parent(self):
-        """Parent Face if assigned. None if not assigned."""
-        return self._parent
-
-    @property
-    def has_parent(self):
-        """Boolean noting whether this Aperture has a parent Face."""
-        return self._parent is not None
-
     def set_adjacency(self, other_aperture):
         """Set this aperture to be adjacent to another.
 
@@ -193,11 +206,11 @@ class Aperture(object):
         """
         assert isinstance(other_aperture, Aperture), \
             'Expected Aperture. Got {}.'.format(type(other_aperture))
-        self.boundary_condition = Surface(other_aperture)
-        other_aperture.boundary_condition = Surface(self)
+        self._boundary_condition = boundary_conditions.surface(other_aperture, True)
+        other_aperture._boundary_condition = boundary_conditions.surface(self, True)
 
-    def overhang(self, depth, angle=0, indoor=False, tolerance=0):
-        """Get a single overhang for this Aperture.
+    def overhang(self, depth, angle=0, indoor=False, tolerance=0, base_name=None):
+        """Add a single overhang for this Aperture.
 
         Args:
             depth: A number for the overhang depth.
@@ -208,18 +221,16 @@ class Aperture(object):
                 indoor geometry). Default: False.
             tolerance: An optional value to return None if the overhang has a length less
                 than the tolerance. Default is 0, which will always yeild an overhang.
-
-        Returns:
-            overhang: A Shade object for the Aperture overhang. You may want to run
-                the check_non_zero() method on the resulting Shade to verify
-                the geometry is subtantial when the Aperture does not have a flat top.
+            base_name: Optional base name for the shade objects. If None, the default
+                is InOverhang or OutOverhang depending on whether indoor is True.
         """
-        overhang = self.louvers_by_number(
-            1, depth, angle=angle, indoor=indoor, tolerance=tolerance)
-        return overhang[0] if len(overhang) != 0 else None
+        if base_name is None:
+            base_name = 'InOverhang' if indoor else 'OutOverhang'
+        self.louvers_by_number(1, depth, angle=angle, indoor=indoor,
+                               tolerance=tolerance, base_name=base_name)
 
-    def right_fin(self, depth, angle=0, indoor=False, tolerance=0):
-        """Get a single vertical fin on the right side of this Aperture.
+    def right_fin(self, depth, angle=0, indoor=False, tolerance=0, base_name=None):
+        """Add a single vertical fin on the right side of this Aperture.
 
         Args:
             depth: A number for the fin depth.
@@ -230,20 +241,16 @@ class Aperture(object):
                 indoor geometry). Default: False.
             tolerance: An optional value to return None if the fin has a length less
                 than the tolerance. Default is 0, which will always yeild a fin.
-
-        Returns:
-            fin: A Shade object for the Aperture fin. You may want to run
-                the check_non_zero() method on the resulting Shade to verify
-                the geometry is subtantial when the Aperture does not have a
-                flat right side.
+            base_name: Optional base name for the shade objects. If None, the default
+                is InRightFin or OutRightFin depending on whether indoor is True.
         """
-        fin = self.louvers_by_number(
-            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
-            indoor=indoor, tolerance=tolerance)
-        return fin[0] if len(fin) != 0 else None
+        if base_name is None:
+            base_name = 'InRightFin' if indoor else 'OutRightFin'
+        self.louvers_by_number(1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
+                               indoor=indoor, tolerance=tolerance, base_name=base_name)
 
-    def left_fin(self, depth, angle=0, indoor=False, tolerance=0):
-        """Get a single vertical fin on the left side of this Aperture.
+    def left_fin(self, depth, angle=0, indoor=False, tolerance=0, base_name=None):
+        """Add a single vertical fin on the left side of this Aperture.
 
         Args:
             depth: A number for the fin depth.
@@ -254,53 +261,56 @@ class Aperture(object):
                 indoor geometry). Default: False.
             tolerance: An optional value to return None if the fin has a length less
                 than the tolerance. Default is 0, which will always yeild a fin.
-
-        Returns:
-            fin: A Shade object for the Aperture fin. You may want to run
-                the check_non_zero() method on the resulting Shade to verify
-                the geometry is subtantial when the Aperture does not have a
-                flat right side.
+            base_name: Optional base name for the shade objects. If None, the default
+                is InLeftFin or OutLeftFin depending on whether indoor is True.
         """
-        fin = self.louvers_by_number(
-            1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
-            flip_start_side=True, indoor=indoor, tolerance=tolerance)
-        return fin[0] if len(fin) != 0 else None
+        if base_name is None:
+            base_name = 'InLeftFin' if indoor else 'OutLeftFin'
+        self.louvers_by_number(1, depth, angle=angle, contour_vector=Vector3D(1, 0, 0),
+                               flip_start_side=True, indoor=indoor,
+                               tolerance=tolerance, base_name=base_name)
 
-    def extruded_border(self, depth, indoor=False):
-        """Get a list of Shade objects from the extruded border of this aperture.
+    def extruded_border(self, depth, indoor=False, base_name=None):
+        """Add a series of Shade objects to this Aperture that form an extruded border.
 
         Args:
             depth: A number for the extrusion depth.
             indoor: Boolean for whether the extrusion should be generated facing the
                 opposite direction of the aperture normal (typically meaning
                 indoor geometry). Default: False.
-
-        Returns:
-            extrusion: A list of Shade objects for the extruded border.
+            base_name: Optional base name for the shade objects. If None, the default
+                is InBorder or OutBorder depending on whether indoor is True.
         """
         extru_vec = self.normal if indoor is False else self.normal.reverse()
         extru_vec = extru_vec * depth
         extrusion = []
         shd_count = 0
+        if base_name is None:
+            shd_name_base = '{}_InBorder{}' if indoor else '{}_OutBorder{}'
+        else:
+            shd_name_base = '{}_' + str(base_name) + '{}'
         for seg in self.geometry.boundary_segments:
             shade_geo = Face3D.from_extrusion(seg, extru_vec)
             extrusion.append(
-                Shade('{}_Shd{}'.format(self.name_original, shd_count), shade_geo))
+                Shade(shd_name_base.format(self.display_name, shd_count), shade_geo))
             shd_count += 1
         if self.geometry.has_holes:
             for hole in self.geometry.hole_segments:
                 for seg in hole:
                     shade_geo = Face3D.from_extrusion(seg, extru_vec)
                     extrusion.append(
-                        Shade('{}_Shd{}'.format(self.name_original, shd_count),
+                        Shade(shd_name_base.format(self.display_name, shd_count),
                               shade_geo))
                     shd_count += 1
-        return extrusion
+        if indoor:
+            self._indoor_shades.extend(extrusion)
+        else:
+            self._outdoor_shades.extend(extrusion)
 
     def louvers_by_number(self, louver_count, depth, offset=0, angle=0,
                           contour_vector=Vector3D(0, 0, 1), flip_start_side=False,
-                          indoor=False, tolerance=0):
-        """Get a list of louvered Shade objects covering this Aperture.
+                          indoor=False, tolerance=0, base_name=None):
+        """Add a series of louvered Shade objects covering this Aperture.
 
         Args:
             louver_count: A positive integer for the number of louvers to generate.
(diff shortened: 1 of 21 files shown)


honeybee/boundarycondition.py
```diff
@@ -21,18 +21,40 @@ def __init__(self, sun_exposure=False, wind_exposure=False,
...
 
+    @classmethod
+    def from_dict(cls, data):
+        """Initialize BoundaryCondition from a dictionary.
+
+        Args:
+            data: A dictionary representaion of the boundary condition.
+        """
+        if 'sun_exposure' in data:
+            sun_exposure = True if data['sun_exposure'] == 'SunExposed' else False
```
mostaphaRoudsari: you probably want to do the comparison for lower(). In EP `sunexposed` or `SUNExposed` are also valid.

honeybee/boundarycondition.py
```diff
@@ -153,6 +175,8 @@ def __init__(self, sun_exposure=True, wind_exposure=True,
                 This input can also be the word 'autocalculate' to have the view
                 factor automatically calculated.  Default: 'autocalculate'.
         """
+        sun_exposure = True if sun_exposure is None else sun_exposure
```
mostaphaRoudsari: similar to what I said above. These checks can be removed.

honeybee/facetype.py
```diff
@@ -6,15 +6,25 @@
...
+        pass
+
+    @classmethod
+    def from_dict(cls, data):
+        """Initialize FaceType from a dictionary.
+
+        Args:
+            data: A dictionary representaion of the face type.
+        """
+        return cls()
```
mostaphaRoudsari: NotImplementedError?

honeybee/model.py
```diff
@@ -22,76 +25,235 @@ class Model(object):
...
+        orphaned_doors
     """
-    __slots__ = ('_name', '_name_original', '_rooms', '_faces', '_shades',
-                 '_apertures', '_doors', '_room_names', '_face_names', '_shade_names',
-                 '_aperture_names', '_door_names', '_properties')
+    __slots__ = ('_name', '_display_name', '_rooms', '_faces', '_shades',
+                 '_apertures', '_doors', '_properties', '_north_angle', '_north_vector')
 
-    def __init__(self, name, objects, north_angle=0):
+    def __init__(self, name, rooms=[], orphaned_faces=[], orphaned_shades=[],
```
mostaphaRoudsari: default values should not be set to empty lists.

Reminder here: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

honeybee/model.py
```diff
@@ -22,76 +25,235 @@ class Model(object):
...
+            'Got {}.'.format(data['type'])
+
+        # import apertures
+        child_apertures = {}
+        orphaned_apertures = []
+        if 'apertures' in data:
+            for ap in data['apertures']:
+                ap_type_class = getattr(hat, ap['aperture_type']['type'])
+                ap_type = ap_type_class.from_dict(ap['aperture_type'])
+                if not ap['boundary_condition']['type'] == 'Surface':
```
mostaphaRoudsari: I can't follow the logic here. Is it supposed to be `if ap['boundary_condition']['type'] == 'Surface':`? now it reads as if it is not surface then set it to outdoors.

honeybee/aperturetype.py
```diff
@@ -5,15 +5,25 @@
...
+        pass
+
+    @classmethod
+    def from_dict(cls, data):
+        """Initialize ApertureType from a dictionary.
+
+        Args:
+            data: A dictionary representaion of the aperture type.
+        """
+        return cls()
```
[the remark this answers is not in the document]
chriswmackey: That one is actually how it's implemented for any aperture type that doesn't have any extra properties on it.  An example of an extra property would be like storing the `fraction_operable` on the `OperableWindow` aperture type.  For such a type with extra properties, we can overwrite this base `from_dict` method such that it correctly sets the `fraction_operable`. So we basically need this simple base method so that we can call `from_dict` for any aperture type, whether it has extra properties or not.

honeybee/boundarycondition.py
```diff
@@ -21,18 +21,40 @@ def __init__(self, sun_exposure=False, wind_exposure=False,
                 This input can also be the word 'autocalculate' to have the view
                 factor automatically calculated.  Default: 'autocalculate'.
         """
+        sun_exposure = False if sun_exposure is None else sun_exposure
```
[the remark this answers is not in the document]
chriswmackey: We don't need it until you implement `from_dict` methods and you realize that these properties are auto-assigned a `None` if they are not included in the dictionary.  In any case, we might just get rid of the `sun_exposure` attributes on the base class and only implement it on the `Outdoors` class depending what we decide in our google doc conversation.  This might simplify the situation here.

honeybee/boundarycondition.py
```diff
@@ -21,18 +21,40 @@ def __init__(self, sun_exposure=False, wind_exposure=False,
...
         assert isinstance(sun_exposure, bool), \
             'Input sun_exposure must be a Boolean. Got {}.'.format(type(sun_exposure))
         self._sun_exposure = sun_exposure
+
+        wind_exposure = False if wind_exposure is None else wind_exposure
```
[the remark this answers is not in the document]
chriswmackey: ditto.

chriswmackey:
@mostaphaRoudsari ,

It looks like most of these comments will be resolved once we discuss the issues in this google doc:
https://docs.google.com/document/d/1lieaQbmHq6Elu6MLaJcQ0g3PjQ40hNNwG60es6HVswM/edit?usp=sharing

Most of the things that you perceive as odd here are the result of schema decisions that we originally thought made sense but have unintended consequences when you play them out.

honeybee/model.py
chriswmackey: Ok. I will add an extra check for `None`, set defaults to `None`, and still permit the setting of empty lists, since that is what will be the case for most models.

honeybee/model.py
chriswmackey: At the moment, Apertures and Doors can only have one of two possible boundary conditions: `Outdoors` and `Surface`.  `Ground` boundary conditions for apertures are illegal as are `Adiabatic` boundary conditions.  If you want, I could just rewrite this statement as:
```
if ap['boundary_condition']['type'] == 'Outdoors':
```
... and it would do the same thing. I was only trying to anticipate a future scenario where we might support another type of boundary condition when I wrote it this way but, now that I think about it, I don't think we will ever support another option for apertures other than those two.  So I can chage it if that's desired.

honeybee/aperturetype.py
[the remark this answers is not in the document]
chriswmackey: This issue no longer exists now that we decided face types and aperture types will just be text strings in the schema.

honeybee/boundarycondition.py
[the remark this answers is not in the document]
chriswmackey: This one no longer exists.

honeybee/boundarycondition.py
chriswmackey: I just made a decision to have `sun_exposure` as a boolean in the schema.  This property is just much better described with a "yes or no" answer that the text values gets unnecessarily confusing, creating more room for errors than they eliminate.  However, I took your suggestion and used a `lower()` check for the `autocalculate` parameter of the `view_factor`.

honeybee/model.py
chriswmackey: This is all now completely re-written.

honeybee/boundarycondition.py
chriswmackey: I just made a decision to have `sun_exposure` as a boolean in the schema.  This property is just much better described with a "yes or no" answer that the text values gets unnecessarily confusing, creating more room for errors than they eliminate.  However, I took your suggestion and used a `lower()` check for the `autocalculate` parameter of the `view_factor`.

honeybee/boundarycondition.py
```diff
@@ -64,57 +120,73 @@ def wind_exposure_idf(self):
...
                 that this boundary condition is assigned.  This other_object will be
                 set as the adjacent object in this boundary condition.
         """
-        _BoundaryCondition.__init__(self)
-        self._boundary_condition_objects = []
+        _boundary_condition_objects = []
         if other_object is not None:
-            self._boundary_condition_objects.append(other_object.name)
+            _boundary_condition_objects.append(other_object.name)
             if other_object.has_parent:
```
mostaphaRoudsari: shouldn't we raise an Exception here? Are you allowing Surface boundary condition for a standalone object?

honeybee/boundarycondition.py
```diff
@@ -64,57 +120,73 @@ def wind_exposure_idf(self):
...
-                self._boundary_condition_objects.append(other_object.parent.name)
+                _boundary_condition_objects.append(other_object.parent.name)
                 if other_object.parent.has_parent:
```
mostaphaRoudsari: This will raise as an error if `other_object.parent` is None

honeybee/door.py
```diff
@@ -73,19 +97,46 @@ def from_vertices(cls, name, vertices, boundary_condition=None):
 
     @property
     def name(self):
-        """The door name (including only legal characters)."""
+        """Get or set the door name (including only legal characters)."""
         return self._name
 
+    @name.setter
```
mostaphaRoudsari: It looks like we can take advantage of sub-classing in this library. We can break down the objects to `Base` and `BaseWithShading` and add name, type, etc to base-class.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
         """A list of all Room objects in the model."""
-        return self._rooms
+        return tuple(self._rooms)
 
     @property
     def faces(self):
         """A list of all Face objects in the model."""
-        return self._faces
+        child_faces = []
+        for room in self._rooms:
```
mostaphaRoudsari: List comprehension is most likely faster for a case like this if we have so many rooms

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
-        running the triangulate_ngon_apertures() method since many models do not
-        have such apertures.
-        """
-        for ap in self.apertures:
-            if len(ap.geometry) > 4:
-                return True
-        return False
+    def orphaned_faces(self):
+        """A list of all Face objects without parent Rooms in the model."""
+        return tuple(self._faces)
```
mostaphaRoudsari: I suggest to keep the naming as `orphaned_faces` and `_orphaned_faces`. I do not have good memories of breaking the rule in naming like this specially now that you have a `faces` property and naturally one would think `_face` corresponds to `faces` and not `orphaned_faces`.

mostaphaRoudsari: Same comment about all the other `orphaned_*` properties.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
 
-    def get_door(self, name):
-        """Get an door object in the model given the door name."""
-        return self._doors[self._door_names.index(name)]
+    @property
+    def orphaned_doors(self):
+        """A list of all Door objects without parent Faces in the model."""
+        return tuple(self._doors)
 
     def add_model(self, other_model):
```
mostaphaRoudsari: Very nice! should we also add `__add__` magic method and use this method under the hood so we can do: `model = model_1 + model_2`

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+        assert not obj.has_parent, 'Door "{}"" has a parent Face. Add the Face to '\
+            'the model instead of the Door.'.format(obj.name)
         self._doors.append(obj)
-        self._door_names.append(obj.name)
+
+    def get_rooms_by_name(self, names):
+        """Get a list of Room objects in the model given the Room names."""
+        rooms = []
+        model_rooms = self._rooms
+        for name in names:
```
mostaphaRoudsari: Not many people use it but Python supports `else` for `for` loops which can simplify cases like this. You can write this loop as:

```
        for name in names:
            for room in model_rooms:
                if room.name == name:
                    rooms.append(room)
                    break
            else:
                raise ValueError('Room "{}" was not found in the model.'.format(name))
```

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+                    break
+            if not found:
+                raise ValueError('Room "{}" was not found in the model.'.format(name))
+        return rooms
+
+    def get_faces_by_name(self, names):
+        """Get a list of Face objects in the model given the Face names."""
+        faces = []
+        model_faces = self.faces
+        for name in names:
```
mostaphaRoudsari: same comment: `for` / `else` can simplify this.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+                    break
+            if not found:
+                raise ValueError('Face "{}" was not found in the model.'.format(name))
+        return faces
+
+    def get_shades_by_name(self, names):
+        """Get a list of Shade objects in the model given the Shade names."""
+        shades = []
+        model_shades = self.shades
+        for name in names:
```
mostaphaRoudsari: same comment: `for` / `else` can simplify this.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+                    break
+            if not found:
+                raise ValueError('Shade "{}" was not found in the model.'.format(name))
+        return shades
+
+    def get_apertures_by_name(self, names):
+        """Get a list of Aperture objects in the model given the Aperture names."""
+        apertures = []
+        model_apertures = self.apertures
+        for name in names:
```
mostaphaRoudsari: same comment: `for` / `else` can simplify this.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+            if not found:
+                raise ValueError(
+                    'Aperture "{}" was not found in the model.'.format(name))
+        return apertures
+
+    def get_doors_by_name(self, names):
+        """Get a list of Door objects in the model given the Door names."""
+        doors = []
+        model_doors = self.doors
+        for name in names:
```
mostaphaRoudsari: same comment: `for` / `else` can simplify this.

honeybee/model.py
```diff
@@ -119,243 +207,391 @@ def north_vector(self, value):
...
+        for door in self._doors:
+            door.scale(factor, origin)
 
     def check_duplicate_room_names(self, raise_exception=True):
-        """Check that there are no duplicate room names in the model."""
-        room_count = len(self._room_names)
-        unique_room_count = len(set(self._room_names))
-        if room_count != unique_room_count:
+        """Check that there are no duplicate Room names in the model."""
+        room_names = []
```
mostaphaRoudsari: You probably want to use `set` in this case which will make the look-up much faster. This is also a good to see:

https://stackoverflow.com/questions/25706136/efficiently-find-repeated-characters-in-a-string/25706298#25706298

mostaphaRoudsari: This comment applies to all check for duplicate items.

honeybee/_shadeparent.py
```diff
@@ -0,0 +1,227 @@
+"""Base class for all geometry objects that can host shades."""
+from .shade import Shade
+
+
+class _ShadeParent(object):
```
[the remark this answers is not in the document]
chriswmackey: Hm.  I could see why you might want to change it but the reality is that the relationship between a Room and its Shades or an Aperture and its Shades is a parent-child object relationship (just like the relationship between Face and Aperture).  Shades can be added/removed from an Aperture in the same way that an Aperture can be added/removed from a Face.
On the other hand, the host/parasite relationship as we have defined it in the rest of the library is essentially an inseparable one from the perspective of the parasite.  The FaceEnergyProperties can only exist when they are attached to a host Face and they cannot be separated from that Face.  So "Parent" is definitely a more appropriate word than "Host" here.  If you want to offer up another naming suggestion, I will consider it but I don't think we should consider making the relationship between shades a true host/parasite relationship, particularly because we want to allow people to make shades with no particular parent.

honeybee/_shadeparent.py
```diff
@@ -0,0 +1,227 @@
...
+    def indoor_shades(self):
+        """Array of all indoor shades assigned to this object."""
+        return tuple(self._indoor_shades)
+
+    @property
+    def shades(self):
+        """Array of all shades (both indoor and outdoor) assigned to this object."""
+        return self._outdoor_shades + self._indoor_shades
+
+    def clear_shades(self):
```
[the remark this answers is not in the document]
chriswmackey: Good point.  I haven't used `remove` in this repo at all but I see that we used it in others so I will change it.

honeybee/_shadeparent.py
```diff
@@ -0,0 +1,227 @@
...
+    def clear_indoor_shades(self):
+        """Remove all indoor shades assigned to this object."""
+        for shade in self._indoor_shades:
+            shade._parent = None
+        self._indoor_shades = []
+
+    def add_outdoor_shade(self, shade):
+        """Add an Shade object to the outdoor of this object.
+
+        Exterior Shade objects can be used to represent overhangs, fins, etc.
```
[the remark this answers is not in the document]
chriswmackey: Thanks for catching this.  I think both "Outdoor" and "Exterior" are fine but "Outdoor" feels a little more generic/abstract and so I think it's slightly better equipped to describe all of the objects that can have outdoor shading.

honeybee/_shadeparent.py
```diff
@@ -0,0 +1,227 @@
...
+            factor: A number representing how much the object should be scaled.
+            origin: A ladybug_geometry Point3D representing the origin from which
+                to scale. If None, it will be scaled from the World origin (0, 0, 0).
+        """
+        for oshd in self._outdoor_shades:
+            oshd.scale(factor, origin)
+        for ishd in self._indoor_shades:
+            ishd.scale(factor, origin)
+
+    def _check_planar_shades(self, tolerance, raise_exception=True):
```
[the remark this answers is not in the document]
chriswmackey: When you run the `check_planar` method over a Room, it checks to see if all geometry across the entire Room is planar.  So this method ends up getting called gets called 3 different times by each of the 3 objects that can have shades as children (Room, Face, Aperture).  Essentially, it saves me from having to write the same code 3 times.

honeybee/aperture.py
```diff
@@ -61,9 +62,41 @@ def __init__(self, name, geometry, type=None, boundary_condition=None):
...
+            boundary_condition = Outdoors.from_dict(data['boundary_condition'])
+        elif data['boundary_condition']['type'] == 'Surface':
+            boundary_condition = Surface.from_dict(data['boundary_condition'])
+        else:
+            raise ValueError(
+                'Boundary condition "{}" is not supported for Apertures.'.format(
+                    data['boundary_condition']['type']))
+        aperture = cls(data['name'], Face3D.from_dict(data['geometry']),
+                       ap_type, boundary_condition)
+        assert aperture.display_name == aperture.name, \
```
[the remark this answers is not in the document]
chriswmackey: This was something that I was going back and forth on.  If the schema is always coming from our Python libraries, I would agree that this is a pointless check.  However, if someone sends us a model JSON they have created on their own and there's an illegal character in the input name, the auto-editing of the name that happens when you initialize the object could cause broken references.  I am thinking specifically about references of adjacent Faces, Apertures, and Doors to one another.  Will all JSONs coming to Pollination Cloud always be run through the validation process when the are received by the server (before they get re-serialized to Python objects)?  If so, I would agree we can remove this check here.

honeybee/aperturetype.py
```diff
@@ -34,18 +32,33 @@ class Window(_ApertureType):
...
+
+        Args:
+            aperture_type_name: A text string for the aperture type (eg. "Window").
+        """
+        attr_name = re.sub('(?<!^)(?=[A-Z])', '_', aperture_type_name).lower()
+        try:
+            return getattr(self, attr_name)
+        except AttributeError:
+            raise AttributeError('Aperture Type "{}" is not supported by this '
+                                 'installation of honeybee.'.format(aperture_type_name))
```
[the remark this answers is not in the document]
chriswmackey: The primary case where I was thinking someone would get this error is when they have created a model using an installation of honeybee with honeybee-energy installed so they assign `GlassDoor` to an Aperture. Then, an installation of honeybee-core without the energy extension goes to re-serialize it and it can't because there's no `GlassDoor` in honeybee-core.

So this is why I thought it was more pertinent to note the installation of honeybee in the message.  Let me know if you still feel we should change it now that I have told you the main case where I expect it to be raised.

honeybee/boundarycondition.py
```diff
@@ -5,18 +5,66 @@
...
+        pass
+
+    @classmethod
+    def from_dict(cls, data):
+        """Initialize BoundaryCondition from a dictionary.
+
+        Args:
+            data: A dictionary representaion of the boundary condition.
+        """
+        return cls()
```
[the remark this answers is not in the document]
chriswmackey: That sounds good to me. So I can put an `assert data[type] == 'Ground'`, then.  And I'll also make sure Ground and Adiabatic get their own from_dict methods.

honeybee/boundarycondition.py
chriswmackey: I think you are right that I have been doing this check in the wrong place.  Right now, I have been doing it in the `set_adjacency` methods on the objects rather than here in the boundary condition.  I originally thought this made sense because, for Apertures, you need to check 2 levels of parents while, for Face, you only need to check 1.  I think I can build it into here by adding an option to include the number of levels to check.

honeybee/boundarycondition.py
chriswmackey: Again, this was previously taken care of in the `set_adjacency` methods but I will move it here.

honeybee/boundarycondition.py
chriswmackey: Done. I think and I think this is better.  People at least get more informative error messages when they set the adjacency for an object that does not have a parent Room.

honeybee/door.py
chriswmackey: `BaseWithShading` sounds much better.  Ok. I'll re-vamp all of the objects now.  It looks like `name` and `geometry` are really the only shared properties but I can still see it saves us a lot of redundant code.

honeybee/_shadeparent.py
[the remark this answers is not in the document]
chriswmackey: Thankfully `BaseWithShade` was a much better choice and we can close this out.

honeybee/_shadeparent.py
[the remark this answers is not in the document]
mostaphaRoudsari: Make sense. Thank you for the explanation.

honeybee/aperture.py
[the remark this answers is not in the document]
mostaphaRoudsari: Yes. All the JSON inputs will be validated by default.

honeybee/aperturetype.py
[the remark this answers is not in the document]
mostaphaRoudsari: I see. We should probably more explicit about this and just include parts of what you said. Something like to support this case ensure that `honeybee-energy` extension is installed.

honeybee/boundarycondition.py
mostaphaRoudsari: Cool!

honeybee/aperture.py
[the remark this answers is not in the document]
chriswmackey: Ok. I will remove these checks, then.

honeybee/aperturetype.py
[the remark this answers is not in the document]
chriswmackey: Hm. Well there isn't really a way to tell if the aperture type was added by `honeybee-energy` or another plugin.  I could add something like this:

```
Aperture Type "{}" is not supported by honeybee. Make sure that you have all necessary honeybee extensions installed'.format(aperture_type_name)
```

Although, the more that I think about it, the more it feels like we are unnecessarily limiting the ability for models to be serialized into a Radiance format just because there is an aperture type that doesn't have much meaning for honeybee-radiance.  It really seems like a shame that we will be unable to set up cloud services with just honeybee-radiance that can re-serialize a model from our plugins just because someone put a glass door in there.  Maybe we really should consider including the other 2 aperture types in the core.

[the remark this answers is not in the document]
chriswmackey: I guess the other option is to just have the type default to `Window` if the type is not recognized.  But it also feels strange to me that you wouldn't want to have this information about whether the Aperture is a Glass Door in honeybee-radiance.  Are you not planning on having something like a `MaterialSet` in honeybee-radiance, which can auto-assign radiance materials depending on face type or aperture type?

honeybee/model.py
chriswmackey: The only reason why I did this previously was because I needed to use `extend` instead of append.  But now I see that it is possible to do stacked list comprehensions like this:
```
[face for room in self._rooms for face in room._faces]
```

honeybee/model.py
chriswmackey: Ok. That is a fair point.

chriswmackey: Changed!

honeybee/model.py
chriswmackey: That does seem very elegant-looking.  I'll add in an `__add__`!

honeybee/model.py
chriswmackey: Very good to know.  Thank you for the tip and I will implement it.

mostaphaRoudsari: Now that you are at it you may want to update this one too:

https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/_lockable.py#L85-L94

Something like this should work:

```python
    if hasattr(cls, '__slots__'):
        for parent_class in getmro(cls)[:-1]:
            if '_locked' in parent_class.__slots__:
                break
        else:
            raise AttributeError(
                'When using the @locakble decorator on a class with __slots__, '
                'a "_locked" variable must be specified within __slots__.')
```

honeybee/aperturetype.py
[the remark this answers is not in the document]
mostaphaRoudsari: I actually just implemented `MaterialSet` which should probably help with this. We can leave it for now and re-address it if it was an issue in the future.

honeybee/model.py
chriswmackey: The main reason why it is implemented this way is because the exact next question that you have after you run this check and get a failure is "what are all of the duplicated names in my model?"  This method allows you to get all of them.

If you think it makes more sense, I can separate each of these into two methods: one that checks whether there are any duplicate names and another that returns a list of all the duplicate names.  I guess this would allow the best of both worlds by allowing for fast performance when the model is good but giving you all of the information you need to fix your model when it is not good.

Also, according to that discussion that you linked to, doesn't using `set` give you worse performance? Isn't using sort or dict better?

mostaphaRoudsari: Your code stays almost the same so you can do all you are doing right now.

```python

    def check_duplicate_room_names(self, raise_exception=True):
        """Check that there are no duplicate Room names in the model."""
        room_names = set()   # this line is changes
        duplicate_names = []
        for room in self._rooms:
            if room.name not in room_names:  # now this is much faster than looking up in a list
                room_names.add(room.name)  # this is also changed
            else:
                duplicate_names.append(room.name)
        if duplicate_names != []:
            if raise_exception:
                raise ValueError('The model has the following duplicated '
                                 'Room names:\n{}'.format('\n'.join(duplicate_names)))
            return False
        return True
```

sort works well with strings or numbers not objects and dict might be too much for what you are trying to do here. Set will give you a much better performance than list for look-ups which will be the main bottleneck here.

mostaphaRoudsari: Also you should probably use set(duplicates) for error otherwise you might get several counts of the same name if the name has been used more than two times.

honeybee/aperturetype.py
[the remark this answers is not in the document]
chriswmackey: Glad to hear that you implemented `MaterialSet`.  That is going to be particularly helpful for making large daylight models and adding urban daylight capabilities to Dragonfly.  And, ok, we can leave everything as it is for now but I have a strong intuition that we will want to move all of the aperture and face types to the core at some point.

honeybee/model.py
chriswmackey: Ok, what you have looks good but, to clarify, we are talking about strings here NOT objects (room names are strings).  So sort might still be the faster option but I also like this implementation with sets.

chriswmackey: Ok, I just implemented it with `set` as you suggested.  Just let me know if you want me to change it for better performance.

mostaphaRoudsari: To be able to use sorted you need to create the list of names first from the Room objects and then use sorted and so on. You can also use sorted with `key` and then compare `.names` but still you are dealing with objects.

chriswmackey: I see what you mean.  Ok, then it seems like `set()` gives us a smaller memory footprint, then, so let's go with that.

chriswmackey:
@mostaphaRoudsari ,

If it's ok with you, I think this is ready to be merged int.

PR #15, opened 2019-08-14
feat(room): Adding indoor_furniture as an alternate to indoor_shade

chriswmackey:


chriswmackey:
BTW, I also decided with this PR to remove some of the places where we had `except Exception` in the `Properties` objects.  I know that these attempts to catch the error were originally well meaning but, more often than not, they stopped me from seeing where the real error was happening within the extension.  In the end, it is better to get the line number where it failed in the extension rather than trying to guess which line it failed on from the exception message.

mostaphaRoudsari:
> I also decided with this PR to remove some of the places where we had except Exception in the Properties objects. I know that these attempts to catch the error were originally well meaning but, more often than not, they stopped me from seeing where the real error was happening within the extension. In the end, it is better to get the line number where it failed in the extension rather than trying to guess which line it failed on from the exception message.

You can use `traceback` in such cases instead of removing the exception. Here is an example:

```python
try:
   print(a)
except NameError as e:
  import traceback
  print(traceback.print_exc())
  raise ValueError('some meaningful error message here!')
```

chriswmackey:
Thanks for the suggestion. Is what you have there exactly what you would suggest (with not importing the module until an exception is thrown)?  You can see in the diff that there are a few of these situations thought the file so I don't know if it makes sense to put the `import` 4 times but I guess if we don't need it until we hit the exception, I can see why it makes sense.

mostaphaRoudsari:
I think either way should work. I have seen a lot of example to import it only on failure which makes sense from performance point of view. You just import it if something fails. At the same time I don't think the overhead is too much if you import it on the top of the module an use it in Exceptions. In both cases it will be imported once.

chriswmackey:
Ok, I just followed your original suggestion and just made one change:
`print(traceback.print_exc())` is redundant since just calling `traceback.print_exc()` already prints the traceback.  Adding `print` just makes a `None` show up in the resulting message.

PR #16, opened 2019-08-16
fix(shade): Fix bug in Shade parent assigning

chriswmackey:


chriswmackey:
Ah, good point @mostaphaRoudsari .  I forgot that we don't have a development branch here and this will break the documentation.  There's no rush to merge this one in so please take your time and do whatever you have to do.  I see @AntoineDao has already pushed a commit, which looks like it should address it.  If not, I am happy to change the tag on my commit from `style` to `fix` as a temporary solution for this one.

mostaphaRoudsari:
We can merge this after we merge the other PR which should address this issue. No need to change the name for the commit.

AntoineDao:
@mostaphaRoudsari I can't actually merge #17 myself as I don't have the permissions to do so. Have a look at the change I've made and merge if you're happy.

chriswmackey:
As we were waiting for the changes on @AntoineDao 's PR, I discovered a pretty serious bug that is affecting my continued development of honeybee-energy.  It was easier to just modify this PR rather than start a new one.  Given this, I might advocate merging this PR in sooner than later, especially now that it will not break the documentation.

PR #17, opened 2019-08-17
ci(docs): build docs in seperate step to PyPi deployment

AntoineDao:
#16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,11 @@ jobs:
     - git config --global user.email "releases@ladybug.tools"
     - git config --global user.name "ladybugbot"
     - npx semantic-release
+  - stage: docs
+    if: branch = master AND (NOT type IN (pull_request))
+    script:
+    - sphinx-apidoc -f -e -d 4 -o ./docs ./ladybug ./ladybug/euclid.py
+    - sphinx-build -b html ./docs ./docs/_build/docs
     deploy:
       provider: pages
       skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 
 install:
   - pip install -r dev-requirements.txt
+  - pip install -r requirements.txt
 
 script:
   - python -m pytest tests/
@@ -24,6 +25,7 @@ jobs:
     - "3.6"
     install:
     - pip install -r dev-requirements.txt
+    - pip install -r requirements.txt
     - npm install @semantic-release/exec
     script:
     - git config --global user.email "releases@ladybug.tools"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+lbt-ladybug==1.6.2
+ladybug-geometry==0.4.3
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -0,0 +1,2 @@
+lbt-ladybug==1.6.2
```
mostaphaRoudsari:
```suggestion
lbt-ladybug==0.4.3
```

requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+ladybug-geometry==0.4.3
```
mostaphaRoudsari: I think you meant to put this as 1.6.2 but we don't need it as `lbt-ladybug` will take care of installing `ladybug-geometry`

.travis.yml
```diff
@@ -24,11 +25,17 @@ jobs:
...
+    - pip install -r requirements.txt
     - npm install @semantic-release/exec
     script:
     - git config --global user.email "releases@ladybug.tools"
     - git config --global user.name "ladybugbot"
     - npx semantic-release
+  - stage: docs
+    if: branch = master AND (NOT type IN (pull_request))
+    script:
+    - sphinx-apidoc -f -e -d 4 -o ./docs ./ladybug ./ladybug/euclid.py
```
mostaphaRoudsari: I should have probably asked this on the other PR too. How does documentation takes care of the new version of the documentation now?

AntoineDao: At the moment it will just overwrite all docs with the new version every time. This is why I mentioned using `readthedocs` on slack.

requirements.txt
AntoineDao: Ah so it does. But ladybug doesn't specify dependency version either: https://github.com/ladybug-tools/ladybug/blob/f4a7afea631f1f6e6e786d12dd4e8c5429d18405/setup.py#L19-L21

AntoineDao: I just merged https://github.com/ladybug-tools/ladybug/pull/155 which addresses the version specific dependency issue.

requirements.txt
AntoineDao: Ah yes you're right my bad. I fixed this in 18af556b7cc242f351e429464a660ac56c54fc38 and also removed ladybug-geometry

.travis.yml
mostaphaRoudsari: but from where does it get the version? we used to pass the version in when running the command (https://github.com/ladybug-tools/honeybee-core/pull/17/files#diff-60254338249f657a0a83f98258a56bfeL13) but how does it know the version when it runs the command here?

requirements.txt
mostaphaRoudsari: Thanks!

.travis.yml
AntoineDao: It doesn't. I can't think of a way to give it a version of we are going to make it rebuild the docs every time it runs deploy on master and also uses semantic release. That is unless we use readthedocs which builds docs based on git tags rather than for every commit.

mostaphaRoudsari: Got it! One more reason to try readthedocs then. Thank you for the quick reply.

PR #19, opened 2019-09-06
Add cli and logger

mostaphaRoudsari:
This PR adds a minimal but extendable Command Line Interface (CLI) to honeybee-core. It also adds documentation for the CLI which will be added to the API documentation.

I will send another PR for `honeybee-radiance` that shows how an extension can take advantage of this set-up to add new commands to honeybee. Meanwhile see the `cli.py`  module docstring.

The PR also adds a logger module for honeybee to set the logging configuration across honeybee libraries. At some point we may want to move the logger to ladybug instead.

The logger has two handlers one to write to honeybee.log file which will log all the logs for review and a stream handler for console which only shows Warning and Error logs. This will probably change as we use the logger more in the core library as well as the extensions.


tests/cli_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+"""
+from click.testing import CliRunner
+from honeybee.cli import viz
+
+def test_viz():
+    runner = CliRunner()
+    result = runner.invoke(viz)
+    assert result.exit_code == 0
+    assert result.output.startswith('vi')
```
AntoineDao: Nice test 😎

docs/index.rst
```diff
@@ -1,16 +1,44 @@
...
+`EnergyPlus <https://energyplus.net/>`_ / `OpenStudio <https://www.openstudio.net/>`_.
+
+Installation
+============
+
+To install the core library try ``pip install -U honeybee-core``.
+
+If you want to also include the command line interface try ``pip install -U honeybee-core[cli]``.
+
+To check if the Honeybee command line is installed correctly try ``honeybee viz`` and you
```
AntoineDao: Does this mean that running `pip install` automatically adds `honeybee` as a terminal/shell command? (if so, super neat!)

setup.py
```diff
@@ -10,16 +9,22 @@
...
     author_email="info@ladybug.tools",
-    description="Honeybee is a Python library to create, run and visualize the results of environmental simulation. See plugins (e.g. honeybee_radiance) for specific simulation type.",
+    description="Honeybee is a Python library to create, run and visualize the results of environmental simulation. See extensions (e.g. honeybee_radiance) for specific simulation type.",
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/ladybug-tools/honeybee-core",
     packages=setuptools.find_packages(exclude=["tests"]),
     install_requires=requirements,
+    extra_requires={
+        'cli': ['click>=5.1']
```
AntoineDao: Is this the section that governs the `pip install honeybee-core[cli]` behaviour?

docs/index.rst
mostaphaRoudsari: yes. it is pretty handy. `pip` creates an script for each  `entry_points` in `console_scripts` in `setup.py`. `honeybee` is mapped to `honeybee.cli.main`.

```python
    entry_points={
        "console_scripts": ["honeybee = honeybee.cli:main"]
    }
```

setup.py
mostaphaRoudsari: yes. the only difference is by adding cli it will install `click`. If you have click installed then `pip install honeybee-core[cli]` and `pip install honeybee-core` will be the same.

I think it will be really useful for us to set the main libraries like `lbt-honeybee` where we can use this tag to install or not install dependencies like `honeybee-radiance`, `honeybee-energy` and so on: `pip install lbt-honeybee[radiance]` and one can always do `pip install lbt-honeybee[all]` to get everything installed.

mostaphaRoudsari:
@AntoineDao thank you for the comments. I think once we have the first workflow ready it will be a good time to chat about logging and the ideal settings.

setup.py
AntoineDao: That is a solid shout. It does mean we should have a bit of a think about dependency management and all of that (as in: plugin architecture is flexible so long as all things are not inter-dependent)

chriswmackey:
Ah wait, you already merged this in?  Ok, maybe we need to do some tests now and revert unless you want to fix them quickly.

chriswmackey:
Yes, I can confirm that you broke the honeybee-core module in ironpython.
image
I realize this isn't quite a sensitive as breaking one of the ladybug modules since honeybee-core really isn't being used by the plugins yet. So I guess we don't have to immediately revert but we probably shouldn't let this be this way for that long (if only to still allow ourselves the ability to test the new honeybee in Rhino when we need to) so let me know how you want to proceed.

AntoineDao:
@chriswmackey I created an issue out of your comments above in #20 so we can keep track of your catch.

PR #21, opened 2019-09-08
fix(logger): write log file to home folder

### LINKED ISSUE #20: v1.7.0 Breaks Grasshopper plugin
Yes, I can confirm that you broke the honeybee-core module in ironpython.
image
I realize this isn't quite a sensitive as breaking one of the ladybug modules since honeybee-core really isn't being used by the plugins yet. So I guess we don't have to immediately revert but we probably shouldn't let this be this way for that long (if only to still allow ourselves the ability to test the new honeybee in Rhino when we need to) so let me know how you want to proceed.

_Originally posted by @chriswmackey in https://github.com/ladybug-tools/honeybee-core/pull/19#issuecomment-529162365_

mostaphaRoudsari:
Fix #20

### DIFF
--- a/honeybee/logutil.py
+++ b/honeybee/logutil.py
@@ -1,29 +1,71 @@
 import logging
 from logging.handlers import TimedRotatingFileHandler
+import os
 
 
-def get_logger(name):
+# This is copied from logging module since python 2 doesn't have it under the same name.
+CRITICAL = 50
+FATAL = CRITICAL
+ERROR = 40
+WARNING = 30
+WARN = WARNING
+INFO = 20
+DEBUG = 10
+NOTSET = 0
+
+_name_to_level = {
+    'CRITICAL': CRITICAL,
+    'FATAL': FATAL,
+    'ERROR': ERROR,
+    'WARN': WARNING,
+    'WARNING': WARNING,
+    'INFO': INFO,
+    'DEBUG': DEBUG,
+    'NOTSET': NOTSET,
+}
+
+
+def _get_log_folder():
+    home_folder = os.getenv('HOME') or os.path.expanduser('~')
+    log_folder = os.path.join(home_folder, '.honeybee')
+    if not os.path.isdir(log_folder):
+        os.mkdir(log_folder)
+    return log_folder
+
+
+def _get_log_level(level):
+    level = _name_to_level.get(level)
+    return level or logging.INFO
+
+def get_logger(name, filename='honeybee.log', file_log_level='DEBUG',
+        console_log_level='WARNING'):
     """Get a logger to be used for each module.
 
-    Use this only once in __init__.py file.
+    Args:
+        name: Logger name. The good practice is to set it to __init__ from inside each
+            modules.
+        filename: Logger filename.Setting filename to None will remove the file handler
+            (Default: honeybee.log).
+        file_log_level: Log level for file handler as a string (Default: DEBUG).
+        console_log_level: Log level for stream handler as a string (Default: WARNING).
     """
     logger = logging.getLogger(name)
-    logger.setLevel(logging.DEBUG)
-
 
     # create a file handler to log debug and higher level logs
-    _file_handler = TimedRotatingFileHandler('honeybee.log', when='midnight')
-    _file_format = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
-    _file_handler.setFormatter(_file_format)
-    _file_handler.setLevel(logging.DEBUG)
+    if filename:
+        log_file = os.path.join(_get_log_folder(), filename)
+        file_handler = TimedRotatingFileHandler(log_file, when='midnight')
+        file_format = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        file_handler.setFormatter(file_format)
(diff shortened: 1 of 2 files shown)


honeybee/logutil.py
```diff
@@ -1,29 +1,42 @@
...
     """Get a logger to be used for each module.
 
-    Use this only once in __init__.py file.
+    Args:
+        name: Logger name. The good practice is to set it to __init__ from inside each
+            modules.
+        filename: Logger filename (Default: honeybee.log).
     """
     logger = logging.getLogger(name)
     logger.setLevel(logging.DEBUG)
```
AntoineDao: Should we not set this to `INFO` and give the option to change the log level to debug when someone is actually trying to debug a script?

honeybee/logutil.py
```diff
@@ -1,29 +1,42 @@
...
-    _stream_handler.setFormatter(_stream_format)
-    _stream_handler.setLevel(logging.WARNING)
+    stream_handler = logging.StreamHandler()
+    stream_format = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
+    stream_handler.setFormatter(stream_format)
+    stream_handler.setLevel(logging.WARNING)
 
-    logger.addHandler(_file_handler)
-    logger.addHandler(_stream_handler)
+    logger.addHandler(file_handler)
```
AntoineDao: Why do we want to send this to a file by default? Would an output in the terminal not suffice? And then one can give the option of stream to a file?

mostaphaRoudsari:
>  My only comment for this PR is that it would be a good idea to note in the docstrings of the cli module that it is not import-able in IronPython (unless my understanding is incorrect here).

I actually don't know! Let me try it out. Click might not be compatible with IronPython but as you said it is not supposed to be imported from IronPython.  I didn't want to put the CLI in a different repo as it will be a pain to keep CLI and the core libraries in sync but if it causes such issues then we may need to rethink that decision. One of the main goals for using the CLI is to free ourselves from running the commands from the command-line outside IronPython but it looks like the IronPython headache won't leave us alone.

honeybee/logutil.py
mostaphaRoudsari: It is very easy to miss the error messages in terminal as we have been experiencing with the plugins right now. Specially with our users who don't know how to code asking them to add a stream to file will be a hard task. I prefer to have them in a file and ask people to look into the file or share the file for debugging.

honeybee/logutil.py
mostaphaRoudsari: We can use INFO but again as I commented down there I'm setting this us based on what a user would do more than a developer.

Also this one doesn't really do anything right now as I'm setting the level for each handler separately. File handler is set to DEBUG and stream handler is set to WARNING.  Are you suggesting to set the File handler to INFO?

honeybee/logutil.py
AntoineDao: That makes sense. In that case can we set it to debug by default but give the option of setting it to something else? I'm thinking of the pollination stuff which most likely won't need the DEBUG level by default.

mostaphaRoudsari:
@chriswmackey thinking about it more one should never import the cli module in the first place. It should be called from the command line so I think we are fine. I will add a note to the module to clarify this. I think I should also remove the modules from the documentation as I did in `honeybee-radiance` so people don't get confused.

honeybee/logutil.py
AntoineDao: That makes sense. In that case can we set it to debug by default but give the option of setting it to something else? I'm thinking of the pollination stuff which most likely won't need the DEBUG level by default.

honeybee/logutil.py
mostaphaRoudsari: Oh OK. I was not clear. The default settings is mainly for developers to be able to debug the code when it fails for non-developer users! 😄 By users here I meant the Grasshopper / Dynamo users who would send this to us for debugging their failed runs. And files are much better than screen-capture in such cases.

honeybee/logutil.py
mostaphaRoudsari: Sure. I will add two input arguments so one can change the log level for each handler when creating the logger.

chriswmackey:
Thanks for clarifying @mostaphaRoudsari and that makes sense.  A note in the documentation would suffice for my concerns, though I guess another option is to make the `cli` module here hidden.  In any case, I agree that this is a situation where practicality beats purity and that the cli module should be included in this repo and not another.

chriswmackey:
Also, it would be nice if that CLI documentation went somewhere that we could reference. If not in the README, then maybe in a wiki page?

mostaphaRoudsari:
With the current settings CLI documentation will be accessible at `https://www.ladybug.tools/[module name]/docs/cli.html`.  For instance for honeybee-core the CLI documentation is here:

https://www.ladybug.tools/honeybee-core/docs/cli.html

What I meant by removing the cli _module_ documentation was to remove this page (https://www.ladybug.tools/honeybee-core/docs/honeybee.cli.html) which after revisiting it is actually helpful for developers to have it in place.

I think the usage of cli will happen is two cases. For normal users it will be fine and for developers they will see the documentation and will be fine after that.

mostaphaRoudsari:
Thank you for your comments. They should be all addressed and this should fix the bug in Grasshopper.

PR #23, opened 2019-09-14
ci(project manager): add initial config

AntoineDao:
This config adds all issues (non pull requests) to a repo and org level kanban boards. Issues will
be moved across boards according to the `labels` value of each column. Finally, new issues and pull requests will have labels automatically added to them as per the config. Some normalisation is also set up to ensure coordinated color between labels.

In order for this to take effect, @mostaphaRoudsari or @chriswmackey will have to authorise the `project-manager` github app to the ladybug-tools org. I have created this app and the source code can be found [here](https://github.com/antoinedao/project-manager).

### DIFF
--- a/.github/project-manager.yml
+++ b/.github/project-manager.yml
@@ -1,0 +1,137 @@
+issue:
+  opened:
+    addLabels:
+      - honeybee
+      - core
+      - triage
+  closed:
+    addLabels:
+      - done
+
+pull_request:
+  opened:
+    addLabels:
+      - honeybee
+      - core
+  closed:
+    addLabels:
+      - done
+
+projects:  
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari:
@AntoineDao should we send this PR to ladybug-tools/honeybee first which has more open issues? I assume that will provide a better option for the initial testing.

chriswmackey:
Is this PR still necessary now that we have merged in [this one](https://github.com/ladybug-tools/honeybee-core/commit/11b4eb1dbdfaea34ca9059e969902189356ed83b)? If not, please close it.

AntoineDao:
Yep totally closable. Good catch Chris thanks :smile:

PR #27, opened 2019-10-13
feat(room): Add property for Room multiplier

chriswmackey:
I know that multipliers clearly have a use in energy simulation but, after thinking about it, they are just as helpful in computing daylight metrics over buildings and so they really should live in honeybee-core. So this commit adds them as a property on the room.

I also took the opportunity to revise several of the docstrings to make it clear which core properties are set-able or not (following a convention that I originally saw in RhinoCommon). You don't really have to review these though, @mostapharoudsari . Just looking over the Room changes and letting me know if you agree is good.

### DIFF
--- a/honeybee/room.py
+++ b/honeybee/room.py
@@ -5,7 +5,7 @@ from .properties import RoomProperties
 from .face import Face
 from .facetype import get_type_from_normal, Wall, Floor
 from .boundarycondition import get_bc_from_position, Outdoors, Surface
-from .typing import float_in_range
+from .typing import float_in_range, int_in_range
 import honeybee.writer as writer
 
 from ladybug_geometry.geometry2d.pointvector import Vector2D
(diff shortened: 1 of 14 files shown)


honeybee/room.py
```diff
@@ -178,9 +181,18 @@ def from_box(cls, name, width=3.0, depth=6.0, height=3.2,
 
     @property
     def faces(self):
-        """Tuple of all honeybee Faces making up this room volume."""
+        """Get a tuple of all honeybee Faces making up this room volume."""
         return self._faces
 
+    @property
+    def multiplier(self):
+        """Get or set an integer noting how many times this Room is repeated."""
```
mostaphaRoudsari: I think it will be helpful to clarify how the room will be repeated which in this case is by being copied vertically in Z direction.

chriswmackey: I can add something here to that nature but I think the general idea of a multiplier is that is just runs the calculation with the room once and just multiplies the result by the multiplier. This means that the "repetition" isn't in any direction (it's essentially in the exact same location).  This comes with some inaccuracy but it can greatly speed up the calculation if the Rooms are similar enough.  Let me know if your understanding is different.

chriswmackey: I put in a more detailed docstring explaining how the multipliers are intended to work.

mostaphaRoudsari: That's true but that is just part of the story. Specially now that we are adding the multiplier to honeybee-core and not honeybee-energy this can change.

For energy modeling the placement of the original zone does make a difference as discussed in this document: https://bigladdersoftware.com/epx/docs/8-9/tips-and-tricks-using-energyplus/using-multipliers-zone-and-or-window.html#background-and-study-using-multipliers

Now think about same model being used for daylighting. What will be  your approach? If we want to see the shade casting effect of the building then we need to copy the geometry around and that needs to be regulated.

PR #28, opened 2019-10-16
fix(to.idf): Remove NotIplementedErrors since to.idf methods are coming

chriswmackey:


chriswmackey:
The tests have to be added within the extension libraries since we clearly shouldn't be testing `to.idf` in the core library. I have tests coming for the `to.idf` methods that I'm adding now in `honeybee-energy`.

PR #31, opened 2019-10-24
feat(search): Add utilities for searching arrays by keyword

chriswmackey:
.. I also fixed some of the tests that were in the previous commit about writers such that these tests will still pass when someone has a honeybee extension installed.

### DIFF
--- a/honeybee/search.py
+++ b/honeybee/search.py
@@ -1,0 +1,51 @@
+"""Collection of methods for searching for keywords and filtering lists by keywords.
+
+This is useful for cases like the following:
+
+    * Searching through the honeybee-radiance modifier library.
+    * Searching through the honeybee-energy material, construction, schedule,
+        constructionset, or programtype libraries.
+    * Searching through EnergyPlus IDD or RDD to find possilbe output variables
+        to request form the simulation.
+"""
+
+
+def filter_array_by_keywords(array, keywords, parse_phrases=True):
+    """Filter an array of strings to get only those containing the given keywords.
+
+    This method is case insensitive, allowing the searching of keywords across
+    different cases of letters.
+
+    Args:
+        array: An array of strings which will be filtered to get only those containing
+            the given keywords.
+        keywords: An array of strings representing keywords.
+        parse_phrases: If True, this method will automatically parse any strings of
+            multiple keywords (spearated by spaces) into separate keywords for
+            searching. This results in a greater liklihood that someone finds what
+            they are searching for in large arrays but it may not be appropropriate for
+            all cases. You may want to set it to False when you are searching for a
+            specific phrase that includes spaces. Default: True.
+    """
+    # split any keywords separated by spaces
+    if parse_phrases:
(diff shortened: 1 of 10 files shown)


honeybee/search.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        array: An array of strings which will be filtered to get only those containing
+            the given keywords.
+        keywords: An array of strings representing keywords.
+    """
+    # split any keywords separated by spaces
+    split_keywords = []
+    for kw in keywords:
+        split_keywords.extend(kw.strip().upper().split(' '))
```
mostaphaRoudsari: This is tricky! What if someone does want to search for the name with space inside? I think we should either give an option for split or let user split the keywords before passing them to this function as an input.

In any case, I would do it slightly different than this to handle cases with more than a single space with no issues (e.g. 'this     keyword'.

```suggestion
        split_keywords.extend(kw.upper().split())
```

Also why don't you use list comprehension in this case?

honeybee/search.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # filter the input array
+    filtered_array = []
+    for item in array:
+        if any_keywords_in_string(item.upper(), split_keywords):
+            filtered_array.append(item)
```
mostaphaRoudsari:
```suggestion
            filtered_array = [item for item in array if any_keywords_in_string(item.upper(), split_keywords)]
```

honeybee/search.py
```diff
@@ -0,0 +1,41 @@
+"""Collection of methods for searching for keywords and filtering lists by keywords."""
```
mostaphaRoudsari: Maybe we should also add in what cases we are using this module.

chriswmackey:
Thanks @mostaphaRoudsari . The main reason why I put it in the core library is because I know we used this method all over honeybee-legacy.  It wasn't just to search for constructions by keywords or search through EnergyPlus output variables but we also used it to filter materials in the radiance library by keyword.  If you're confident that you don't want to have this capability for radiance materials in the [+] plugin, I can move it out of the core. However, if you think you will make use of it in Radiance workflows, I think it's similar to the typing module in that we should keep it in the core.

honeybee/search.py
chriswmackey: We never had this case in honeybee-legacy (this splitting method was always the preferred one).  Still, I will add an option to not split it for the cases people might need the space in the search.

List comprehension is always a fine substitute for looping and appending be we need to use extend here. I can do a for loop within a for loop using list comprehension here but that is usually where I draw the line and just say that looping and extending is just as elegant.  I'll switch it to list comprehension since I see that you have a stronger preference for it.

chriswmackey: Now that you mentioned it, I just thought of one of the new cases where I want to apply this method and I realized it's one where including the space is important. So I wholeheartedly agree that we should expose it as an option.

honeybee/search.py
chriswmackey: Done!

PR #32, opened 2019-10-25
feat(adjacency): Output adjacency information from solve_adjacency

chriswmackey:
After deliberating a bit on different ways to do this, it seems best to just return the information about the adjacencies that were found during the solve_adjacency run. This way, the information can be used to assign custom properties to the newly-adjacent faces, apertures and doors. For example, assigning custom interior materials/constructions or setting all new adjacent faces to be air walls.

### DIFF
--- a/honeybee/face.py
+++ b/honeybee/face.py
@@ -384,43 +384,58 @@ class Face(_BaseWithShade):
             other_face: Another Face object to be set adjacent to this one.
             tolerance: The minimum distance between the center of two aperture
                 geometries at which they are condsidered adjacent. Default: 0.
+
+        Returns:
+            adjacent_apertures -- A list of tuples with each tuple containing 2 objects
+                for Apertures paired in the process of solving adjacency.
+            adjacent_doors -- A list of tuples with each tuple containing 2 objects
+                for Doors paired in the process of solving adjacency.
         """
         # check the inputs and the ability of the faces to be adjacent
         assert isinstance(other_face, Face), \
             'Expected honeybee Face. Got {}.'.format(type(other_face))
+
         # set the boundary conditions of the faces
         self._boundary_condition = boundary_conditions.surface(other_face)
         other_face._boundary_condition = boundary_conditions.surface(self)
+
         # set the apertures to be adjacent to one another
         assert len(self._apertures) == len(other_face._apertures), \
             'Number of apertures does not match between {} and {}.'.format(
                 self.name, other_face.name)
+        adjacent_apertures = []
         if len(self._apertures) > 0:
             found_adjacencies = 0
             for aper_1 in self._apertures:
                 for aper_2 in other_face._apertures:
                     if aper_1.center.distance_to_point(aper_2.center) <= tolerance:
                         aper_1.set_adjacency(aper_2)
+                        adjacent_apertures.append((aper_1, aper_2))
                         found_adjacencies += 1
                         break
             assert len(self._apertures) == found_adjacencies, \
                 'Not all apertures of {} were found to be adjacent to apertures in {}.' \
                 '\nTry increasing the tolerance.'.format(self.name, other_face.name)
+
         # set the doors to be adjacent to one another
         assert len(self._doors) == len(other_face._doors), \
             'Number of doors does not match between {} and {}.'.format(
                 self.name, other_face.name)
+        adjacent_doors = []
         if len(self._doors) > 0:
             found_adjacencies = 0
             for door_1 in self._doors:
                 for door_2 in other_face._doors:
                     if door_1.center.distance_to_point(door_2.center) <= tolerance:
                         door_1.set_adjacency(door_2)
+                        adjacent_doors.append((door_1, door_2))
                         found_adjacencies += 1
                         break
             assert len(self._doors) == found_adjacencies, \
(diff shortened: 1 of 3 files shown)


honeybee/face.py
```diff
@@ -384,44 +384,59 @@ def set_adjacency(self, other_face, tolerance=0):
             other_face: Another Face object to be set adjacent to this one.
             tolerance: The minimum distance between the center of two aperture
                 geometries at which they are condsidered adjacent. Default: 0.
+
+        Returns:
+            adjacent_apertures -- A list of tuples with each tuple containing 2 objects
```
mostaphaRoudsari: any reason we don't use the same structure as args?
```suggestion
            adjacent_apertures: A list of tuples with each tuple containing 2 objects.
```

honeybee/room.py
```diff
@@ -558,11 +575,16 @@ def solve_adjacency(rooms, tolerance):
                             if not isinstance(face_2.boundary_condition, Surface):
                                 if face_1.geometry.is_centered_adjacent(
                                         face_2.geometry, tolerance):
-                                    face_1.set_adjacency(face_2)
```
mostaphaRoudsari: I see what are you trying to do here and from your explanation it is about giving better access to the objects but I feel there should be a better approach to do this.

To me it looks much cleaner to set the adjacency as you have it here. I would say one can access the objects and change construction, etc. afterwards.

honeybee/face.py
chriswmackey: Yes, there's a good reason. Sphinx isn't able to render the returns in the same structure as the Args but the double dash gives some clarity as you can see in the ladybug docs hre:
https://www.ladybug.tools/ladybug/docs/ladybug.datacollection.html#ladybug.datacollection.DailyCollection.filter_collections_by_statement

honeybee/room.py
chriswmackey: Figuring out which adjacencies were solved for in the run of `solve_adjacnecy` and which ones were already existing before the run isn't easy without some output for this method. So I'm afraid the alternative you propose here isn't really usable without some major caveats. Here are some of the things I considered as an alternative to what is implemented here:

1. In the Grasshopper component, just don't use this staticmethod and write my own method to solve adjacency inside the component that sets custom constructions and face types at the time when we are setting adjacency between faces. I think you'd agree that this is not good and it's much better to have a method in the Python libraries that we actually use.

2. In the Grasshopper component, loop through all of the Rooms before running `solve_adjacnecy` to find Faces that already have an adjacent Face. Then run this `solve_adjacnecy` method and again loop through all of the Faces in an attempt to determine which ones are new. Assign custom constructions to just the new ones. I think you will agree that the current implementation of just returning the new adjacency information is much cleaner than having to loop through the data twice and attempt to figure out what happened.

3. Add inputs to this `solve_adjacency` staticmethod for custom interior radiance materials and constructions, even though this is the core library and it is supposed to be agnostic towards particular extensions. For this, I'd have to add a try/except loop in the staticmethod that handles the case that a particular extension is not installed. I think you'd agree that this is in violation of our vision to leave extension-specific stuff out of the core library. If we make an exception here for particular extensions, then people might make an argument to include code about their own extensions down the line. In the words of Python Zen "Special cases aren't special enough to break the rules."

If you think any of these proposals is better than what I have implemented, let me know. Also, let me know if there's another alternative that I am not seeing.

honeybee/face.py
mostaphaRoudsari: I see. In that case I suggest to follow something like this one instead of coming up with a new syntax: https://www.sphinx-doc.org/en/1.5/ext/example_google.html

honeybee/room.py
```diff
@@ -558,11 +575,16 @@ def solve_adjacency(rooms, tolerance):
...
+                                    adj_aps, adj_drs = face_1.set_adjacency(face_2)
+                                    adjacent_faces.append((face_1, face_2))
+                                    adjacent_apertures.extend(adj_aps)
+                                    adjacent_doors.extend(adj_drs)
                                     break
             except IndexError:
                 pass  # we have reached the end of the list of zones
 
+        return adjacent_faces, adjacent_apertures, adjacent_doors
```
mostaphaRoudsari: @chriswmackey this is a generic comment about how to write methods / functions and if it is fine to return multiple values. I don't have the correct answer but I know this will be an issue in the long-term.

From Python point-of-view this is totally fine as you are returning a tuple. From my experience when you move to returning multiple values from a method it doesn't scale well and will make breaking changes in the future.

What will happen, and we had this for the legacy several times, is for an upcoming change we will need to add or remove an output to a method. This will potentially break all the code where this method is called. A possible solution is to enforce returning a Named Tuple of a Dictionary with key / values. That means adding a new output won't break the current use cases. Dictionaries will add an overhead but in the long term it makes the maintenance and development easier.

Related: https://stackoverflow.com/a/61636/4394669

chriswmackey: I would agree that this is a case where a dictionary is best much like what we do for [comfort functions n ladybug-comfort](https://github.com/ladybug-tools/ladybug-comfort/blob/master/ladybug_comfort/pmv.py#L88-L96). Let's agree that it's worth considering using this dictionary approach any time there are more than 4 outputs being returned or there's even a slight chance that the method may output more variables in the future.

PR #39, opened 2019-11-02
feat(rename): Add rename method to all objects

chriswmackey:
This method will ease the cases where someone duplicates and edits a starting object and then wants to combine it with the original object into one Model (like making a model of repeated rooms) since all objects within a Model must have unique names.

### DIFF
--- a/honeybee/_base.py
+++ b/honeybee/_base.py
@@ -47,7 +47,7 @@ class _Base(object):
         """Get object properties, including Radiance, Energy and other properties."""
         return self._properties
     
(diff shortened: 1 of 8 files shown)


chriswmackey:
Thanks @mostaphaRoudsari .  This is why I asked you to review this one.  I'll change the name to `add_prefix` and I'll add a separator but I'm reserving `..` for some special internal cases where we may want to split the name later.  For this one, I think I'll just use `_` and, if someone needs a separator for their own workflows, they can include it within the prefix.

PR #45, opened 2019-12-12
feat(orientation): Add module to handle orientation subdivision

chriswmackey:
This new module constitutes all of the generic functions for assigning different properties based on Face or Aperture orientation.  I imagine that these will get applied all over honeybee and its extensions given how useful this type of automation is. And I know @mostapharoudsari mentioned in an earlier review that he felt these functions belonged in the core rather than repeated over several components.

### DIFF
--- a/honeybee/door.py
+++ b/honeybee/door.py
@@ -5,6 +5,7 @@ from .properties import DoorProperties
 from .boundarycondition import boundary_conditions, Outdoors, Surface
 import honeybee.writer.door as writer
 
+from ladybug_geometry.geometry2d.pointvector import Vector2D
 from ladybug_geometry.geometry3d.pointvector import Point3D
 from ladybug_geometry.geometry3d.face import Face3D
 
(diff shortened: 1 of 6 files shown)


honeybee/door.py
```diff
@@ -194,6 +195,37 @@ def perimeter(self):
...
+
+        Args:
+            north_vector: A ladybug_geometry Vector2D for the north direction.
+                Default is the Y-axis (0, 1).
+        """
+        return math.degrees(
+            north_vector.angle_clockwise(Vector2D(self.normal.x, self.normal.y)))
+
+    def cardinal_direction(self, north_vector=Vector2D(0, 1)):
+        """Get text description for the cardinal direction that the door is pointing.
```
mostaphaRoudsari: I know it is named cardinal but I would go for 8 or 16 orientations if that makes sense. Not sure about the exact use case but that will be more flexible.

chriswmackey: Ok. I'll make this change for all of the objects that have this method.

PR #46, opened 2019-12-22
feat(config): Add module and json to handle configurations

chriswmackey:
The config file handles several commonly-accessed folders including the folders for the default simulation folder and the location where the python exe is installed.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include honeybee/config.json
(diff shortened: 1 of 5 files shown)


chriswmackey:
@mostaphroudsari , I am going to just merge this one in before review since this commit is following all of the suggestions that you made over on the honeybee_energy PR and I need this PR merged to get the tests to pass over on honeybee_energy.  If you find anything in this PR that you would like changed, just let me know and I will send an edit.

PR #52, opened 2020-01-27
feat(shade): Add property to track whether Shade is on the indoors

chriswmackey:
This is_indoor property is needed to correctly set up the ModifierSet on honeybee-radiance.

Also, as part of this commit, I am adding the ability to assign Shades to Doors. This has now come up enough times from more than one person that I've realized that I better implement it now while I still have a near-complete idea of all the other repositories affected by the change (another month or two and I don't think that I would be able to catch all of the places that will be incorrect as a result of this change). As far as I can tell, the following repos need to be changed with a PR that parallels this one:
* honeybee-energy (especially the Model objectand Model writer)
* honeybee-schema (including both Door schema and the Wiki)
* energy-model-measure (OpenStudio has no concept of assigning Shades to Doors but I'll figure something out)
* honeybee-grasshopper-core (most of the visualization components and  the component that assigns Shade to a parent)
* honeybee-grasshopper-energy (the components that assign constructions and schedules to shades)

Lastly, I added a new method to the Face to add repeating windows at a fixed width x height now that we have methods for this on ladybug_geometry. This doesn't require any parallel PRs in other repos but, at some point, I should probably add a WindowParameter object to Dragonfly that makes use of this new method.

### DIFF
--- a/honeybee/face.py
+++ b/honeybee/face.py
@@ -482,8 +482,11 @@ class Face(_BaseWithShade):
                 rectangle is made. It does not affect the ability to produce apertures.
 
         Usage:
+
+        .. code-block:: python
+
             room = Room.from_box(3.0, 6.0, 3.2, 180)
-            room[1].apertures_by_ratio(0.4)
+            room.faces[1].apertures_by_ratio(0.4)
         """
         assert 0 <= ratio < 1, 'Ratio must be between 0 and 1. Got {}'.format(ratio)
         self._acceptable_sub_face_check(Aperture)
@@ -530,8 +533,11 @@ class Face(_BaseWithShade):
                 considered a part of a rectangle.
 
         Usage:
+
+        .. code-block:: python
+
             room = Room.from_box(3.0, 6.0, 3.2, 180)
-            room[1].apertures_by_ratio_rectangle(0.4, 2, 0.9, 3)
+            room.faces[1].apertures_by_ratio_rectangle(0.4, 2, 0.9, 3)
         """
         assert 0 <= ratio <= 0.95, \
             'Ratio must be between 0 and 0.95. Got {}'.format(ratio)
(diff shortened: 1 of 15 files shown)


honeybee/face.py
```diff
@@ -548,14 +548,65 @@ def apertures_by_ratio_rectangle(self, ratio, aperture_height, sill_height,
...
+                the aperture_height will take precedence.
+            horizontal_separation: A number for the target separation between
+                individual apertures centerlines.  If this number is larger than
+                the parent rectangle base, only one aperture will be produced.
+            tolerance: The maximum difference between point values for them to be
+                considered a part of a rectangle.
+
+        Usage:
+            room = Room.from_box(3.0, 6.0, 3.2, 180)
+            room[1].apertures_by_ratio_rectangle(0.4, 2, 0.9, 3)
```
mostaphaRoudsari: This is not related to this PR but just seeing this I feel we should use `room.faces[1]` in our sample code which is more explicit way to access the faces.

honeybee/face.py
```diff
@@ -548,14 +548,65 @@ def apertures_by_ratio_rectangle(self, ratio, aperture_height, sill_height,
             aperture._parent = self
             self._apertures.append(aperture)
 
+    def apertures_by_width_height_rectangle(self, aperture_height, aperture_width,
```
mostaphaRoudsari: Maybe this needs to have the word repeating somewhere. It is easy to miss the `s` in the call. I couldn't come up with a good short name and `repeating_apertures_by_width_height_rectangle` is too long. Just an observation.

honeybee/face.py
chriswmackey: Ok. I'm glad that you pointed this out since I also just realised that this usage example is not correct (it seems that I forgot to change it after copy/pasting it from another docstring). I'll fix it now.

honeybee/face.py
chriswmackey: I'm wondering if it makes sense to use the word `repeating` instead of `rectangle`. Basically, I want to communicate that, just like `apertures_by_ratio_rectangle`, this `apertures_by_width_height_rectangle` method will fill any rectangular portions identified within the `Face3D` with repeating rectangular windows. Let me think about this since a naming change like this should happen to both of these methods and will require PRs to a few others that are making use of them. If we eventually decide that we want to change it, I'll send another round of PRs.

honeybee/face.py
chriswmackey: Done!

PR #53, opened 2020-02-03
fix(boundarycondition): Update based on recent schema change

chriswmackey:
This commit brings the library up-to-date with these recent changes in honeybee-schema:
https://github.com/ladybug-tools/honeybee-schema/pull/44

Only the to_dict methods have been changed and the Python objects still internally use strings for simplicity. An alternative approach could use classes for AutoCalculate, AutoSize and NoLimit and use them similarly to how the `facetype` module is used. A change like this would involve a bunch of edits in a lot of different places, though. @mostapharoudsari, what do you think?

### DIFF
--- a/honeybee/altnumber.py
+++ b/honeybee/altnumber.py
@@ -1,0 +1,47 @@
+"""Objects used as alternatives to various numerical properties."""
+
+
+class _AltNumber(object):
+    __slots__ = ()
+
+    def __init__(self):
+        pass
+
+    @property
+    def name(self):
+        return self.__class__.__name__
+    
+    def to_dict(self):
+        """Get the object as a dictionary."""
+        return {'type': self.name}
+
+    def ToString(self):
+        return self.__repr__()
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return self.name
+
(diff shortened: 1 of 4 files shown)


chriswmackey:
@mostaphaRoudsari . I think, unfortunately, this isn't the type of issue that we can easily say that "we'll implement it later" because it introduces breaking changes across almost all of the libraries and plugins that are only going to multiply with time. To put it another way, trying to make a change like this in the future is going to take 10 times as long as making the change now. So I think it might be more helpful to know if you think using Python objects as opposed to strings is an improvement over using strings or not.

In one sense, classes are much easier to check than strings, they can have their own to_dict and from_dict that save us some work on other classes, and they can have a `value` property that helps us translate the object to a string for EnergyPlus.  On the other hand, using objects will probably make it more difficult for new developers since a `honeybee_energy` developer needs to know to import the `AutoSize` class from `honeybee` core in order to set a lot of the energy objects to autosize.

If you think classes are still an improvement, let's implement it now before I add more objects that need to use the autosize feature of EnergyPlus.

chriswmackey:
Thinking it over, the benefits of having objects seem to outweigh the one downside, which is probably a rare occurrence anyway given that `Autosize` and `Autocalculate` tend to be defaults.  I guess it also seems weird to me to have an object in our schema that isn't in our Python libraries. I'll make the change now.

PR #57, opened 2020-02-07
fix(default): Do not use mutable objects for input argument defaults

chriswmackey:


honeybee/extensionutil.py
```diff
@@ -2,8 +2,8 @@
 """A series of utility functions that are useful across several honeybee extensions."""
 
 
-def model_extension_dicts(data, extension_key, room_ext_dicts=[], face_ext_dicts=[],
-                          shade_ext_dicts=[], aperture_ext_dicts=[], door_ext_dicts=[]):
+def model_extension_dicts(data, extension_key, room_ext_dicts, face_ext_dicts,
```
mostaphaRoudsari: I would have set them to None instead of making them required. Then you could have set it to an empty list inside the function. I assume that would have made it easier for extensions which are going to take advantage of this function.

chriswmackey: Using `None` is what I would do in most normal situations. But these functions get called potentially thousands of times when re-serializing a Model (essentially getting called for each and every re-serialized geometry object). This just adds up to a lot of extra `if`/`else` statements to handle cases where `None` is input. And realistically, I'm not going to use the `None` option given that these functions are really only used once in each extension. For other more commonly-used functions, though, I will use `None`

PR #60, opened 2020-02-11
Format docstrings for correct display in sphinx-generated bootstrap pages 

santiagogaray:
Hi @chriswmackey 

This time I found these error messages when I generated the pages.  

honeybee - ERROR - Failed to import honeybee_energy!
Traceback (most recent call last):
  File "C:\Users\sgara\Documents\GitHub\honeybee-core\honeybee\__init__.py", line 16, in <module>
    extensions[name] = importlib.import_module(name)
  File "c:\python27\lib\importlib\__init__.py", line 37, in import_module
    __import__(name)
  File "c:\python27\lib\site-packages\honeybee_energy\__init__.py", line 8, in <module>
    import honeybee_energy._extend_honeybee
  File "c:\python27\lib\site-packages\honeybee_energy\_extend_honeybee.py", line 19, in <module>
    from .properties.model import ModelEnergyProperties
  File "c:\python27\lib\site-packages\honeybee_energy\properties\model.py", line 5, in <module>
    from ..lib.constructionsets import generic_construction_set
  File "c:\python27\lib\site-packages\honeybee_energy\lib\constructionsets.py", line 2, in <module>
    from honeybee_energy.constructionset import ConstructionSet
  File "c:\python27\lib\site-packages\honeybee_energy\constructionset.py", line 5, in <module>
    from .material.opaque import EnergyMaterial, EnergyMaterialNoMass
  File "c:\python27\lib\site-packages\honeybee_energy\material\opaque.py", line 10, in <module>
    from ..writer import generate_idf_string
  File "c:\python27\lib\site-packages\honeybee_energy\writer.py", line 8, in <module>
    from honeybee.facetype import RoofCeiling, AirWall
ImportError: cannot import name AirWall

I believe I installed all the dependencies specified there in the repo and everything seems to be okay when I navigated the pages. Is this something to ignore?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -100,11 +100,17 @@ html_theme_options = {
     # Fix navigation bar to top of page?
     # Values: "true" (default) or "false"
     'navbar_fixed_top': "true",
-	'navbar_pagenav': True,
+    'navbar_pagenav': True,
     'source_link_position': "nav",
-	'bootswatch_theme': "united",
+    'bootswatch_theme': "united",
     'bootstrap_version': "3",
-	}
+}
+
+# Bootstrap theme custom file paths (relative to this file)
+# Layout.html path (already added above, include if different)
+# templates_path = ['_templates']
+# Stylesheet path
+html_static_path = ["_static"]
 
 # on_rtd is whether we are on readthedocs.org
 # on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
@@ -117,7 +123,7 @@ html_theme_options = {
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = []
+# html_static_path = []
 
 # Custom sidebar templates, must be a dictionary that maps document names
 # to template names.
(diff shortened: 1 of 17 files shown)


chriswmackey:
Your pip-installed version of `honeybee-energy` is out of sync with this version of `honeybee-core`. I would just remove it since it will inject extra things into your local documentation if you have it installed.  Just type:
```
pip uninstall honeybee-energy
```
Select `y`, then you should be able to build the docs locally without issues.

santiagogaray:
Thanks @chriswmackey. I did exactly that and everything run well as you anticipated.
I did check that format-wise everything looks good, I believe. Feel free to take a look at it  and let me know if you have any comments when you get a chance.

docs/conf.py
```diff
@@ -117,7 +123,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = []
+# html_static_path = []
```
chriswmackey: Is there a reason why this is commented out here?
I can see that it's not commented out in the ladybug_tools_template:
https://github.com/ladybug-tools/ladybug-tools-template/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/docs/conf.py#L114

Or wait, maybe it is?
https://github.com/ladybug-tools/ladybug-tools-template/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/docs/conf.py#L127

Why do we have both a commented and un-commented version of this line? Could they just be a single un-commented line? Sorry that I didn't catch this in earlier PRs.

santiagogaray: Good pont @chriswmackey.  Actually, I did look at this same thing today and I was wondering what was the the logic I sued at the time to leave it like this. My take is that many things were changed and tested out and ended up as a leftover. 
My suggestion is that we should uncomment the second path and delete the top one. I can do this to the ladybug_tools_template. Actually I can go back to all the other repos to be consistent if you think makes sense.

santiagogaray:
> BTW, I think the [`honeybee-energy` repo](https://github.com/ladybug-tools/honeybee-energy) will be a good next target. It's a big one so brace yourself!

Whoohoo!! its gonna be fun!!!

chriswmackey:
Thanks, @santiagogaray . This looks good and let's merge it!

PR #67, opened 2020-04-02
refactor(identifier): Change name property to identifier

chriswmackey:
This is the first of a series of commits to 12 repos that change the name property to uuid.

This PR should NOT be merged in until comparable PRs can be put together for the other 12 repos that are broken by this change.

### DIFF
--- a/honeybee/_base.py
+++ b/honeybee/_base.py
@@ -7,52 +7,78 @@ class _Base(object):
     """A base class for all geometry objects.
 
     Args:
-        name: Object name. Must be < 100 characters.
+        identifier: Text string for a unique object ID. Must be < 100 characters and
+            not contain any spaces or special characters.
 
     Properties:
-        * name
+        * identifier
         * display_name
+        * user_data
     """
-    __slots__ = ('_name', '_display_name', '_properties')
+    __slots__ = ('_identifier', '_display_name', '_properties', '_user_data')
 
-    def __init__(self, name):
+    def __init__(self, identifier):
         """Initialize base object."""
-        self.name = name
+        self._identifier = valid_string(identifier, 'honeybee object identifier')
+        self._display_name = self._identifier
         self._properties = None
+        self._user_data = None
 
     @property
-    def name(self):
-        """Get or set the object name (including only legal characters)."""
-        return self._name
-
-    @name.setter
-    def name(self, value):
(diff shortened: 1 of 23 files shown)


honeybee/_base.py
```diff
@@ -7,40 +7,38 @@ class _Base(object):
...
-        be the same. Legal characters are ., A-Z, a-z, 0-9, _ and -.
-        Invalid characters are automatically removed from the original name for
-        compatability with simulation engines.
+        If not set by the user, this will be equal to the uuid.
         """
         return self._display_name
 
+    @display_name.setter
+    def display_name(self, value):
+        self._display_name = str(value)
```
mostaphaRoudsari, line 40: We should test this and see how it will work with non-ASCII characters.

chriswmackey: I will add a test for it but I think the point of the display_name is that it can use non-ASCII characters since we don't use the display_name in any simulation engines.

chriswmackey: Hm. I see that everything seems to be OK in Python 3 with regards to non-ASCII characters but the Python 2 json module doesn't seem to be able to handle the re-serialization:
https://travis-ci.org/github/ladybug-tools/honeybee-core/builds/670273420?utm_source=github_status&utm_medium=notification

Let me see if there's any way around this within the Python 2 json module and if not, I guess we will have to strip out non-ASCII characters.

chriswmackey: Actually, I realized that the issue isn't with the json module. It's that the Python 2.7 file didn't compile on Travis while there's a non-ascii character within the Python file. So this just seems to be a matter of the Travis machine lacking the character set. I can see that assigning the non-ascii character is working all well and good from Rhino and that the Travis machine with Python 2.7 can still load up a JSON file with a non-ascii character in it (it's just that it only uses the code for the character while the machine lacks the character set). I feel that, if we don't provide a property free from ascii restrictions now, we will want to include one at some point in the future (I just can't imagine us being used internationally without some way to store characters of other languages). So I think we should not restrict the display_name for now.

What do you think?

chriswmackey: Now that I added an try/except to catch Unicode errors, I think the `display_name` is all safe to accept non-ascii characters. I'll make sure that all of the display_name properties have this.

PR #74, opened 2020-04-08
build(deps-dev): bump dependency versions

mostaphaRoudsari:
The version change is breaking a number of tests. @chriswmackey I didn't get into the details here as it will probably be much easier for you to figure out. I assume it is because the updates in the latest version of schema but I can be wrong.


mostaphaRoudsari:
Never mind! It passed all tests.

Issue #106, opened 2020-05-12
Add a story_identifier property to honeybee Room


chriswmackey:
Given that many OpenStudio measures expect this, I think we should just add it as a property to the Room.

We can also add a staticmethod to the Room class to auto-assign stories based on floor height (overwriting any stories with `None` story_identifiers).  We can also expose this as a method on the Model class and call it on the Grasshopper end of things before exporting the model to OpenStudio.

PR #121, opened 2020-05-21
feat(room): Add story property to Room

chriswmackey:
This commit adds a story property to honeybee-core Rooms and add a staticmethod to auto-assign stories based on floor heights.

Resolves https://github.com/ladybug-tools/honeybee-core/issues/106

### DIFF
--- a/honeybee/model.py
+++ b/honeybee/model.py
@@ -62,13 +62,16 @@ class Model(_Base):
         * angle_tolerance
         * rooms
         * faces
-        * shades
         * apertures
         * doors
+        * shades
+        * indoor_shades
+        * outdoor_shades
         * orphaned_faces
         * orphaned_shades
         * orphaned_apertures
         * orphaned_doors
+        * stories
         * user_data
     """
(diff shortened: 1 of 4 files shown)


chriswmackey:
@mostaphaRoudsari , I think I will just merge this so that I can finish the integration of this with dragonfly. I think much of the code here shouldn't be that far from what you were expecting but, if you have any comments or suggestions, just let me know and I can send another PR.

honeybee/model.py
```diff
@@ -373,6 +376,18 @@ def orphaned_shades(self):
...
+    def stories(self):
+        """Get a list of text for each unique story identifier in the Model.
+
+        Note that this will be an empty list if the model has to rooms.
+        """
+        _stories = set()
+        for room in self._rooms:
+            if room.story is not None:
+                _stories.add(room.story)
+        return list(_stories)
```
mostaphaRoudsari, line 389: I suggest to sort the list before returning it. Otherwise you can get them in a different order every time.

PR #195, opened 2020-08-14
feat(typing): Add 3 new float range checks to typing.

saeranv:
@chriswmackey 

Here are the 3 new float_in_range checks I added to check variable inclusive/exclusive ranges, required for the AFN.

Sorry for not assigning reviewers, and assignees, looks like I don't have access to this repo.

Let me know how it looks. 

S

### DIFF
--- a/honeybee/typing.py
+++ b/honeybee/typing.py
@@ -71,13 +71,19 @@ def valid_ep_string(value, input_name=''):
     return value
 
 
-def float_in_range(value, mi=INFNEG, ma=INFPOS, input_name=''):
-    """Check a float value to be between minimum and maximum."""
+def _number_check(value, input_name):
+    """Check if value is a number."""
     try:
         number = float(value)
     except (ValueError, TypeError):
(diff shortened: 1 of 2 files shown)


honeybee/typing.py
```diff
@@ -83,6 +83,42 @@ def float_in_range(value, mi=INFNEG, ma=INFPOS, input_name=''):
     return number
 
 
+def float_in_range_excl(value, mi=INFNEG, ma=INFPOS, input_name=''):
+    """Check a float value to be greater than minimum and less than maximum."""
+    try:
+        number = float(value)
+    except (ValueError, TypeError):
+        raise TypeError('Input {} must be a number. Got {}: {}.'.format(
+            input_name, type(value), value))
```
chriswmackey, line 92: Rather than repeating these same 5 lines of code 6 times, can we just pull it out into its own hidden method?  You can just call it `_number_check` and just call it from each of the other places here.

saeranv:
@chriswmackey, great, I just added the new function.

chriswmackey:
Looks good. Merged!

PR #239, opened 2020-10-14
feat(cli): Add groups for creating and editing models

chriswmackey:
This commit adds two new groups for creating models  and editing their geometry.

The create group houses methods for creating model JSONs for typical geometry cases like shoe boxes, rectangles and (eventually), regular polygons, L and H-shaped plans, etc.

The edit group houses methods for solving adjacency, converting units, generating windows by ratio, generating overhangs.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 honeybee-schema==1.39.0;python_version>='3.7'
+ladybug-geometry-polyskel==1.3.11
 coverage==5.3
 coveralls==1.7.0;python_version<'3.0'
 coveralls==2.1.2;python_version>='3.6'
--- a/honeybee/cli/create.py
+++ b/honeybee/cli/create.py
@@ -1,0 +1,207 @@
+"""honeybee model creation commands."""
+
+try:
+    import click
+except ImportError:
+    raise ImportError(
+        'click is not installed. Try `pip install . [cli]` command.'
+    )
+
+from ladybug_geometry.geometry2d.pointvector import Point2D, Vector2D
+from ladybug_geometry.geometry2d.polygon import Polygon2D
+from ladybug_geometry.geometry3d.pointvector import Point3D, Vector3D
+from ladybug_geometry.geometry3d.face import Face3D
+from ladybug_geometry.geometry3d.polyface import Polyface3D
+from ladybug_geometry_polyskel.polysplit import perimeter_core_subpolygons
+
+from honeybee.model import Model
+from honeybee.room import Room
+from honeybee.facetype import face_types as fts
+from honeybee.boundarycondition import boundary_conditions as bcs
+try:
+    ad_bc = bcs.adiabatic
+except AttributeError:  # honeybee_energy is not loaded and adiabatic does not exist
+    ad_bc = None
+
+import sys
+import os
+import logging
+import json
+import math
+import uuid
+
+_logger = logging.getLogger(__name__)
+
+
+@click.group(help='Commands for creating Honeybee models.')
+def create():
+    pass
+
+
+@create.command('shoe-box')
+@click.argument('width', type=float)
+@click.argument('depth', type=float)
+@click.argument('height', type=float)
+@click.option('--orientation-angle', '-a', help='A number between 0 and 360 for the '
+              'clockwise orientation of the box in degrees. (0=North, 90=East, 180='
+              'South, 270=West).', type=float, default=0, show_default=True)
+@click.option('--window-ratio', '-wr', help='A number between 0 and 1 (but not equal to 1) '
+              'for the ratio between aperture area and area of the face pointing towards'
+              ' the orientation-angle.', type=float, default=0.4, show_default=True)
+@click.option('--adiabatic/--outdoors', ' /-o', help='Flag to note whether the faces that '
+              'are not in the direction of the orientation-angle are adiabatic or '
+              'outdoors.', default=True, show_default=True)
+@click.option('--units', '-u', help=' Text for the units system in which the model '
+              'geometry exists. Must be (Meters, Millimeters, Feet, Inches, Centimeters).',
+              type=str, default='Meters', show_default=True)
+@click.option('--tolerance', '-t', help='The maximum difference between x, y, and z '
+              'values at which vertices are considered equivalent.',
+              type=float, default=0.01, show_default=True)
+@click.option('--output-file', '-f', help='Optional file to output the Model JSON string.'
+              ' By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def shoe_box(width, depth, height, orientation_angle, window_ratio, adiabatic,
+             units, tolerance, output_file):
+    """Create a model with a single shoe box Room.
+    \n
+    Args:\n
+        width: Number for the width of the box (in the X direction).\n
+        depth: Number for the depth of the box (in the Y direction).\n
+        height: Number for the height of the box (in the Z direction).
+    """
+    try:
+        unique_id = str(uuid.uuid4())[:8]  # unique identifier for the shoe box
+
+        # create the box room and assign all of the attributes
+        room_id = 'Shoe_Box_Room_{}'.format(unique_id)
+        room = Room.from_box(room_id, width, depth, height, orientation_angle)
+        room.display_name = 'Shoe_Box_Room'
+        front_face = room[1]
+        front_face.apertures_by_ratio(window_ratio, tolerance)
+        if adiabatic and ad_bc:
+            room[0].boundary_condition = ad_bc  # make the floor adiabatic
+            for face in room[2:]:  # make all other face adiabatic
+                face.boundary_condition = ad_bc
+
+        # create the model object
+        model_id = 'Shoe_Box_Model_{}'.format(unique_id)
+        model = Model(model_id, [room], units=units,
+                      tolerance=tolerance, angle_tolerance=1)
+
+        # write the model out to the file or stdout
+        output_file.write(json.dumps(model.to_dict()))
+    except Exception as e:
+        _logger.exception('Shoe box creation failed.\n{}'.format(e))
+        sys.exit(1)
+    else:
+        sys.exit(0)
+
+
+@create.command('rectangle-plan')
+@click.argument('width', type=float)
+@click.argument('length', type=float)
+@click.argument('floor-to-floor-height', type=float)
+@click.option('--perimeter-offset', '-p', help='An optional positive number that will be'
+              ' used  to offset the perimeter to create core/perimeter Rooms. '
+              'If this value is 0, no offset will occur and each floor will have one '
+              'Room', type=float, default=0, show_default=True)
+@click.option('--story-count', '-s', help='An integer for the number of stories to '
+              'generate.', type=int, default=1, show_default=True)
+@click.option('--orientation-angle', '-a', help='A number between 0 and 360 for the '
+              'clockwise orientation that the width of the box faces.(0=North, 90=East, '
+              '180=South, 270=West).', type=float, default=0, show_default=True)
+@click.option('--outdoor-roof/--adiabatic-roof', ' /-ar', help='Flag to note whether '
+              'the roof faces of the top floor should be outdoor or adiabatic.',
+              default=True, show_default=True)
+@click.option('--ground-floor/--adiabatic-floor', ' /-af', help='Flag to note whether '
+              'the floor faces of the bottom floor should be ground or adiabatic.',
+              default=True, show_default=True)
+@click.option('--units', '-u', help=' Text for the units system in which the model '
+              'geometry exists. Must be (Meters, Millimeters, Feet, Inches, Centimeters).',
+              type=str, default='Meters', show_default=True)
+@click.option('--tolerance', '-t', help='The maximum difference between x, y, and z '
+              'values at which vertices are considered equivalent.',
+              type=float, default=0.01, show_default=True)
+@click.option('--output-file', '-f', help='Optional file to output the Model JSON '
+              'string. By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def rectangle_plan(width, length, floor_to_floor_height, perimeter_offset, story_count,
+                   orientation_angle, outdoor_roof, ground_floor, units, tolerance,
+                   output_file):
+    """Create a model with a rectangular floor plan.\n
+    Note that the resulting Rooms in the model won't have any windows or solved
+    adjacencies and the edit commands should be used for this purpose.\n
+    \n
+    Args:\n
+        width: Number for the width of the plan (in the X direction).\n
+        depth: Number for the depth of the plan (in the Y direction).\n
+        floor_to_floor_height: Number for the height of each floor of the model
+            (in the Z direction).
+    """
+    try:
+        unique_id = str(uuid.uuid4())[:8]  # unique identifier for the rooms
+
+        # create the geometry of the rooms for the first floor
+        footprint = Face3D.from_rectangle(width, length)
+        if perimeter_offset != 0:  # use the straight skeleton methods
+            assert perimeter_offset > 0, 'perimeter_offset cannot be less than than 0.'
+            try:
+                footprint = []
+                base = Polygon2D.from_rectangle(Point2D(), Vector2D(0, 1), width, length)
+                sub_polys_perim, sub_polys_core = perimeter_core_subpolygons(
+                    polygon=base, distance=perimeter_offset, tol=tolerance)
+                for s_poly in sub_polys_perim + sub_polys_core:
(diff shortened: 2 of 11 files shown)


dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
 honeybee-schema==1.39.0;python_version>='3.7'
+ladybug-geometry-polyskel==1.3.11
```
mostaphaRoudsari, line 2: Shouldn't this b e a requirement for anyone who wants to run the commands from the CLI?

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@create.command('shoe-box')
+@click.argument('width', type=float)
+@click.argument('depth', type=float)
+@click.argument('height', type=float)
+@click.option('--orientation-angle', help='A number between 0 and 360 for the clockwise'
+              'orientation of the box in degrees. (0 = North, 90 = East, 180 = South, '
+              '270 = West).', type=float, default=0, show_default=True)
+@click.option('--window-ratio', help='A number between 0 and 1 (but not equal to 1) '
```
mostaphaRoudsari, line 48: Can we also add sill height to give more control in case one wants to do daylight simulation?

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+              'for the ratio between aperture area and area of the face pointing towards'
+              ' the orientation-angle.', type=float, default=0.4, show_default=True)
+@click.option('--adiabatic/--outdoors', help='Flag to note whether the faces that are '
+              'not in the direction of the orientation-angle are adiabatic or outdoors.',
+              default=True, show_default=True)
```
mostaphaRoudsari, line 53:
```suggestion
              default=True, flag=True, show_default=True)
```

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+@click.option('--units', help=' Text for the units system in which the model geometry '
+              'exists. Choose from (Meters, Millimeters, Feet, Inches, Centimeters).',
+              type=str, default='Meters', show_default=True)
+@click.option('--tolerance', help='The maximum difference between x, y, and z values '
+              'at which vertices are considered equivalent.',
+              type=float, default=0.01, show_default=True)
+@click.option('--output-file', help='Optional file to output the Model JSON string. '
+              'By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def shoe_box(width, depth, height, orientation_angle, window_ratio, adiabatic,
```
mostaphaRoudsari, line 63: I can see that you have a separate command to add overhangs but I feel one of the most requested option will be adding horizontal/vertical shades by size and distance. Similar to the one that we support in Grasshopper.

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+              'If this value is 0, no offset will occur and each floor will have one '
+              'Room', type=float, default=0, show_default=True)
+@click.option('--story-count', help='An integer for the number of stories to generate.',
+              type=int, default=1, show_default=True)
+@click.option('--orientation-angle', help='A number between 0 and 360 for the clockwise'
+              'orientation that the width of the box is facing.(0 = North, 90 = East, '
+              '180 = South, 270 = West).', type=float, default=0, show_default=True)
+@click.option('--outdoor-roof/--adiabatic-roof', help='Flag to note whether the roof '
+              'faces of the top floor should be outdoor or adiabatic.',
+              default=True, show_default=True)
```
mostaphaRoudsari, line 115:
```suggestion
              default=True, flag=True, show_default=True)
```

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+@click.option('--ground-floor/--adiabatic-floor', help='Flag to note whether the floor '
+              'faces of the bottom floor should be ground or adiabatic.',
+              default=True, show_default=True)
```
mostaphaRoudsari, line 118:
```suggestion
              default=True, flag=True, show_default=True)
```

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+              type=float, default=0.01, show_default=True)
+@click.option('--output-file', help='Optional file to output the Model JSON string. '
+              'By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def rectangle_plan(width, length, floor_to_floor_height, perimeter_offset, story_count,
+                   orientation_angle, outdoor_roof, ground_floor, units, tolerance,
+                   output_file):
+    """Create a model with a rectangular floor plan.\n
+    Note that the resulting Rooms in the model won't have any windows or solved
+    adjacencies and the edit commands should be used for this purpose.\n
```
mostaphaRoudsari, line 133: Hmmm. Again I think we should support such features in this command. What stops you from at least doing the solve adjacency. Or I should say in what scenario do you imagine one will not want to do the solve adjacency and how common is that?

honeybee/cli/create.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # create the model object
+        model_id = 'Rectangle_Plan_Model_{}'.format(unique_id)
+        model = Model(model_id, rooms, units=units,
+                      tolerance=tolerance, angle_tolerance=1)
+
+        # write the model out to the file or stdout
+        output_file.write(json.dumps(model.to_dict()))
+    except Exception as e:
+        _logger.exception('Shoe box creation failed.\n{}'.format(e))
```
mostaphaRoudsari, line 204:
```suggestion
        _logger.exception('Rectangle model creation failed.\n{}'.format(e))
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+@click.group(help='Commands for editing Honeybee models.')
+def edit():
+    pass
+
+
+@edit.command('convert-to-units')
+@click.argument('model-json', type=click.Path(
+    exists=True, file_okay=True, dir_okay=False, resolve_path=True))
+@click.argument('units', type=str)
+@click.option('--output-file', help='Optional file to output the Model JSON string'
```
mostaphaRoudsari, line 40: Can we have the option just to change the units and not re-size the geometry? Like how Rhino asks if one wants to re-size the model when the units are changed.

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+              ' with solved adjacency. By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def solve_adjacency(model_json, units, output_file):
```
mostaphaRoudsari, line 43:
```suggestion
def convert_units(model_json, units, output_file):
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+    else:
+        sys.exit(0)
+
+
+@edit.command('solve-adjacency')
+@click.argument('model-json', type=click.Path(
+    exists=True, file_okay=True, dir_okay=False, resolve_path=True))
+@click.option('--wall/--air-boundary', help='Flag to note whether the '
+              'wall adjacencies should be of the air boundary face type.',
+              default=True, show_default=True)
```
mostaphaRoudsari, line 75:
```suggestion
              default=True, flag=True, show_default=True)
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+@click.option('--surface/--adiabatic', help='Flag to note whether the adjacencies '
+              'should be surface or adiabatic.', default=True, show_default=True)
```
mostaphaRoudsari, line 77:
```suggestion
              'should be surface or adiabatic.', default=True, flag=True, show_default=True)
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+        _logger.exception('Model solve adjacency failed.\n{}'.format(e))
+        sys.exit(1)
+    else:
+        sys.exit(0)
+
+
+@edit.command('windows-by-ratio')
+@click.argument('model-json', type=click.Path(
+    exists=True, file_okay=True, dir_okay=False, resolve_path=True))
+@click.argument('ratio', type=float)
```
mostaphaRoudsari, line 126: Can we pass multiple ratios similar to ratio by orientation?

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+        sill_height: A number for the target height above the bottom edge of
+            the rectangle to start the apertures. Note that, if the
+            ratio is too large for the height, the ratio will take precedence
+            and the sill_height will be smaller than this value.
+        horizontal_separation: A number for the target separation between
+            individual aperture centerlines.  If this number is larger than
+            the parent rectangle base, only one aperture will be produced.
+    """
+    try:
+        assert os.path.isfile(model_json), 'No JSON file found at {}.'.format(model_json)
```
mostaphaRoudsari, line 204: You don't need this check. `Path(exists=True)` does this check for you.

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+    exists=True, file_okay=True, dir_okay=False, resolve_path=True))
+@click.argument('depth', type=float)
+@click.option('--angle', help='A number for the for an angle to rotate the overhang '
+              'in degrees. Positive numbers indicate a downward rotation while '
+              'negative numbers indicate an upward rotation.',
+              type=float, default=0, show_default=True)
+@click.option('--vertical-offset', help='An optional number for the vertical offset of '
+              'the overhang from the top of the window or face. Positive numbers move '
+              'up while negative mode down.', type=float, default=0, show_default=True)
+@click.option('--by-window/--by-wall', help='Flag to note whether the overhangs should '
```
mostaphaRoudsari, line 243: I would suggest `per-window` and `per-wall`. It will be more explicit.

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+              'be generated per aperture or per wall.', default=True, show_default=True)
```
mostaphaRoudsari, line 244:
```suggestion
              'be generated per aperture or per wall.', default=True, flag=True, show_default=True)
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+@click.option('--outdoor/--indoor', help='Flag to note whether the overhangs should be '
+              'on the indoors like a light shelf.', default=True, show_default=True)
```
mostaphaRoudsari, line 246:
```suggestion
              'on the indoors like a light shelf.', default=True, flag=True, show_default=True)
```

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+              type=click.File('w'), default='-')
+def overhang(model_json, depth, angle, vertical_offset, by_window, outdoor, output_file):
+    """Add overhangs to all outdoor walls or windows in walls.\n
+    \n
+    Args:
+        model_json: Full path to a Model JSON file.\n
+        depth: A number for the overhang depth.
+    """
+    try:
+        assert os.path.isfile(model_json), 'No JSON file found at {}.'.format(model_json)
```
mostaphaRoudsari, line 258: no need for this check. similar to the one before.

honeybee/cli/edit.py
```diff
@@ -0,0 +1,292 @@
...
+              'on the indoors like a light shelf.', default=True, show_default=True)
+@click.option('--output-file', help='Optional file to output the Model JSON string'
+              ' with overhangs. By default it will be printed out to stdout',
+              type=click.File('w'), default='-')
+def overhang(model_json, depth, angle, vertical_offset, by_window, outdoor, output_file):
+    """Add overhangs to all outdoor walls or windows in walls.\n
+    \n
+    Args:
+        model_json: Full path to a Model JSON file.\n
+        depth: A number for the overhang depth.
```
mostaphaRoudsari, line 255: What about adding multiple shades?

mostaphaRoudsari (commented):
Looks good! I have some questions about how do you imagine adding multiple shades and also about streamlining the process by merging the steps.

There are also a number of minor bugs and possible improvements.

Thank you for putting all of this together quickly!

dev-requirements.txt
chriswmackey: Yes, you will see it is in `setup.py` under the `[cli]` option.

honeybee/cli/create.py
chriswmackey: Yes, there's a separate command that gives control over all that.

honeybee/cli/create.py
chriswmackey: So you don't need the `flag=True` when you write the option the way that I did with the `/`.
https://click.palletsprojects.com/en/7.x/options/#boolean-flags

To me, it seemed more elegant and explicit to use the `/` but let me know if you redundancy of the `flag=True` is helpful

honeybee/cli/create.py
chriswmackey: Yes, similarity to grasshopper is why it's in a separate command. The idea is that you can generate the base geometry with the command here and then pass it through some of the shade or window-generating commands, which have tons of customization options like Grasshopper.  I am breaking things down so that I don't get a single command that has 30+ inputs.

honeybee/cli/create.py
chriswmackey: I don't want to create a command with 30+ inputs.  You see that the solve adjacency command already has 5 options. Moving those over to here isn't just going to create a lot of copy/pasted code.  It's going to mean that this command has almost 15 inputs.

honeybee/cli/edit.py
chriswmackey: Yes, we can add a flag for that option.

honeybee/cli/edit.py
chriswmackey: Good catch. I think my flake8 filter has been flaking on me.

honeybee/cli/edit.py
chriswmackey: We should definitely support that functionality at some point but I think that it should probably be through a separate command or through some way that doesn't make this command more difficult to use.  And we have to agree on some convention for inputting lists into the commands.  Is a JSON string of a list the way you were thinking of implementing it?

honeybee/cli/edit.py
chriswmackey: Ah, very good point. I forgot that I have this relic from before I used all that the click decorator offers.  I'll fix it in a few places now.

chriswmackey: Done!

honeybee/cli/edit.py
chriswmackey: Sounds good.  Will do.

chriswmackey: Done

honeybee/cli/edit.py
chriswmackey: I will add two separate commands for generating louvers by count and louvers by distance between. I just wanted to get your thoughts on this one first before adding them.

honeybee/cli/create.py
mostaphaRoudsari: Nice! Didn't know about that.

honeybee/cli/edit.py
chriswmackey: Added!

honeybee/cli/edit.py
mostaphaRoudsari: You can pass multiple value for an option by using multiple=True.

https://click.palletsprojects.com/en/7.x/options/#multiple-options

I think we can include it in this command.

honeybee/cli/create.py
mostaphaRoudsari: I'm used to Radiance so 15 options is not really something that I would mind but I can see that it can be potentially confusing for people. Let's keep them separate for now and see how it will turn out in the future. We can always add another command that does it all in one go.

honeybee/cli/edit.py
chriswmackey: But that is for a `click.option` and not a `click.argument`.  Are you suggesting that we just make the ratio optional with a default?

Also, how will queenbee interact with lists of inputs in this format?

mostaphaRoudsari: You can make ratio an option with a default value.

> Also, how will queenbee interact with lists of inputs in this format?

Queenbee supports ArrayInput which is a list of numbers. Let's keep the commands for honeybee clean and resolve any possible issues on the Queenbee side.

chriswmackey: I guess it's just that trying to add this list functionality really feels un-clean to me. As does changing some of these arguments into options (especially things like `aperture_height` for which a default value is invariably going to be making a potentially incorrect assumption about the units of the model).  In other words, I feel like I'm only adding this list functionality for the sake of queenbee because I'm assuming that queenbee has some more elegant way of inputting lists.

PR #272, opened 2020-11-11
feat(tojson): Module to write json or hbjson from Honeybee objects

devngc:
This PR is created in response to [this](https://github.com/ladybug-tools/honeybee-3dm/pull/2#discussion_r520663891) comment from Mostapha.

### DIFF
--- a/honeybee/tojson.py
+++ b/honeybee/tojson.py
@@ -1,0 +1,57 @@
+# coding: utf-8
+"""Write json or hbjson from Honeybee objects"""
+
+# Import standard python libraries
+import os
+import json
+
+# import the core honeybee dependencies
+from .model import Model
+from .config import folders
+
+
+def to_json(hb_objects, name=None, folder_path=None, indent=None, abridged=False):
+    """This function write Honeybee objects to json file.
+
+    Args:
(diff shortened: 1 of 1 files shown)


chriswmackey:
@devngc ,
I would say that a method like this is too implicit for the core libraries as it is not clear which objects it is trying to address.  For example, Models don't have the option of abridged vs not abridged (so that input is meaningless for that case) yet they also have options that no other objects have in their `to_dict` methods, like the option to triangulate sub faces.

So this really seems like it should be UI code for a case where we don't know what object we are serializing.  And, for all core libraries, your JSON writing should be explicit and sensitive to the object type you are translating.  After all, it only takes 2 lines of Python to turn a `to_dict` method into a `to_json` one.

Let's see if @mostaphaRoudsari agrees.

mostaphaRoudsari (commented):
Maybe I was not clear in my note but by implementing it to core libraries I meant adding methods to hand `to_json` not a new module. For instance we can add a `to_json` method to Model which calls `to_dict()` under the hood and writes it to a JSON file.

Then you can call `Model.to_json` from inside the 3dm library when needed.

PR #274, opened 2020-11-11
feat(model): A method to write json

devngc:


honeybee/model.py
```diff
@@ -1349,6 +1375,21 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
 
         return base
 
+    @staticmethod
+    def to_json(hb_dict, folder_path, name="unnamed"):
```
chriswmackey, line 1379: The arguments here should match those of the `to_dict` method on the model object but with a couple of new ones specifically for the JSON serialization.  I think a single `file_path` argument is probably better than separate arguments for folder and name. It's probably also good to include an `indent` argument as this is pretty useful.

honeybee/model.py
```diff
@@ -1349,6 +1375,21 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+        """Writes a Honeybee model dictionary to hbjson
+
+        Args:
+            hb_dict (dictionary): A Honeybee model in the form of a dictionary
+            folder_path (A text string): Folder location where json will be written
+            name (str, optional): File name. Defaults to "unnamed".
+        """
+        file_name = '{}.hbjson'.format(name)
```
chriswmackey, line 1387: What if the name already ends with `.hbjson`?  The Wea.write() method is probably a good example for you to follow:
https://github.com/ladybug-tools/ladybug/blob/master/ladybug/wea.py#L968

honeybee/model.py
```diff
@@ -1349,6 +1375,21 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
         return base
 
+    @staticmethod
+    def to_json(hb_dict, folder_path, name="unnamed"):
+        """Writes a Honeybee model dictionary to hbjson
+
+        Args:
+            hb_dict (dictionary): A Honeybee model in the form of a dictionary
+            folder_path (A text string): Folder location where json will be written
+            name (str, optional): File name. Defaults to "unnamed".
```
chriswmackey, line 1385: These are not formatted correctly for sphinx and won't show up correctly in the online docs.  The argument type should come after the `:`. See all of the other docstrings in this file.  You should also probably read the contributing docs on this:
https://github.com/ladybug-tools/contributing/wiki/1.-How-to-write-docstrings-for-Ladybug-Tools

chriswmackey (changes requested):
Thanks @devngc .
I agree that a `to_json` method on the Model class is the right way to do it. But there are several changes that need to be made before it's good to go.

honeybee/model.py
```diff
@@ -141,17 +142,20 @@ def from_dict(cls, data):
             rooms = [Room.from_dict(r, tol, angle_tol) for r in data['rooms']]
         orphaned_faces = None  # import orphaned faces
         if 'orphaned_faces' in data and data['orphaned_faces'] is not None:
-            orphaned_faces = [Face.from_dict(f) for f in data['orphaned_faces']]
+            orphaned_faces = [Face.from_dict(f)
+                              for f in data['orphaned_faces']]
```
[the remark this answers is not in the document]
devngc: Hi @chriswmackey, @mostaphaRoudsari,
Sorry for these formatting issues. I have line length set at 89 characters. I am not sure why this happens. I also have my ruler set at line length 89
settings

devngc:
Hi @chriswmackey, I believe I was able to address all your comments except one about having `name` as an argument to the method. I believe we need this argument on the method or else how can user name the HBJSON? 

If we want to avoid the name argument then we can use either a default name "unnamed" or we can use the Honeybee model's display_name. If the user has not assigned a name to the Honeybee model then the model name also becomes "unnamed". Threfore, I believe it's good to have this argument on this new instance method that we're adding to the model module.

The new PR is [here](https://github.com/ladybug-tools/honeybee-core/pull/277)

honeybee/model.py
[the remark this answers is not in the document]
devngc: I believe I figured why this was happening. No response needed.

PR #277, opened 2020-11-12
feat(to_hbjson): Write hbjson from a Honeybee model

devngc:


devngc:
@chriswmackey,
This is ready for review.
Thanks!

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+                Apertures and Doors) should be triangulated if they have more than
+                4 sides (True) or whether they should be left as they are (False).
+                This triangulation is necessary when exporting directly to EnergyPlus
+                since it cannot accept sub-faces with more than 4 vertices. Note that
+                setting this to True will only triangulate sub-faces with parent Faces
+                that also have parent Rooms since orphaned Apertures and Faces are
+                not relevant for energy simulation. Default: False.
+        """
+        # Create dictionary from the Honeybee Model
+        hb_dict = self.to_dict(included_prop=None, triangulate_sub_faces=False)
```
mostaphaRoudsari, line 1376:
```suggestion
        hb_dict = self.to_dict(included_prop=included_prop, triangulate_sub_faces=triangulate_sub_faces)
```

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+
+        # Setting up a name for the HBJSON
+        if not name.lower().endswith('.hbjson'):
+            file_name = '{}.hbjson'.format(name)
+        else:
+            name = name[:-7]
```
mostaphaRoudsari, line 1382: I'm not sure why this is needed here?

mostaphaRoudsari (commented):
I added two comments while we are waiting for @chriswmackey!

honeybee/model.py
devngc: This is in case a user gives a file name of ,"testhbjson".

mostaphaRoudsari: Hmm... why don't you just use `test.hbjson`? You are removing the extension in this line and then adding it back again in next line?

devngc: I get your point. I am removing `.hbjson` from the name and adding back in. Whereas I could just use that as a name. I think I will change that.

mostaphaRoudsari (approved):
LGTM! Thanks.

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
 
         return base
 
+    def to_hbjson(self, name="unnamed", folder_path=None, indent=None, included_prop=None, triangulate_sub_faces=False):
```
chriswmackey, line 1354: This is longer than 89 characters.

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+        """Writes a Honeybee model to HBJSON.
+
+        Args:
+            name: A text string that will be the name of the HBJSOn. Defaults to "unnamed".
```
chriswmackey, line 1358: `n` of `HBJSON` should be uppercase. The same below.

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+            folder_path: A text string of path to folder where HBJSOn will be written. Defaults to None.
```
chriswmackey, line 1359: You should explain what the `None` means. Same below.

honeybee/model.py
```diff
@@ -1349,6 +1351,46 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+        """
+        # Create dictionary from the Honeybee Model
+        hb_dict = self.to_dict(included_prop=included_prop,
+                               triangulate_sub_faces=triangulate_sub_faces)
+
+        # Setting up a name for the HBJSON
+        if name.lower().endswith('.hbjson'):
+            file_name = name
+        else:
+            file_name = '{}.hbjson'.format(name)
```
chriswmackey, line 1383: You can do this more elegantly with a couple of lines:
```
file_name  = name if name.lower().endswith('.hbjson') or name.lower().endswith('.json') \
    else '{}.hbjson'.format(name)
```

honeybee/model.py
chriswmackey: Also, can the default indent just be 0?

chriswmackey (commented):
Generally looks good. I just had a few comments.

chriswmackey:
And one more comment:
We could really use a test for this code.  Maybe you can include it in the existing `Model.to_dict()` method.

honeybee/model.py
devngc: Revised

honeybee/model.py
devngc: Revised

honeybee/model.py
devngc: Explanation added

honeybee/model.py
devngc: Revised

devngc:
@chriswmackey, A new test is added in `model_test` module for the the `to_hbjson` method

honeybee/model.py
```diff
@@ -1349,6 +1351,50 @@ def to_dict(self, included_prop=None, triangulate_sub_faces=False):
...
+                  included_prop=None, triangulate_sub_faces=False):
+        """Writes a Honeybee model to HBJSON.
+
+        Args:
+            name: A text string that will be the name of the HBJSON.
+                Defaults to "unnamed" for the file name for HBJSON.
+            folder_path: A text string of path to folder where HBJSON will be written.
+                Defaults to None. If folder_path is not specified, the default simulation
+                folder will be used to write the HBJSON. This default simulation folder
+                is at "C:\\Users\\USERNAME\\simulation."
```
chriswmackey, line 1364: This is alright but, for future reference, this is only for Windows machines.  It's different on Mac and Linux.  So it's probably better not to state it in the docstring.

chriswmackey (approved):
Looks good. Let's merge.

honeybee/model.py
devngc: Noted

PR #287, opened 2020-11-25
fix bug in to_hbjson

devngc:
Give out a message if the Honeybee model is not found as an input to this method.

### DIFF
--- a/honeybee/model.py
+++ b/honeybee/model.py
@@ -3,6 +3,7 @@
 from __future__ import division
 import os
 import json
+import warnings
 
 from ._base import _Base
 from .checkdup import check_duplicate_identifiers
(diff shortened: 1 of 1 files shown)


devngc:
I am not sure why the test for Python 2.7 is failing. I tried a similar code block on Python 2.7. It runs without errors.

chriswmackey:
It's failing because you changed the version of importlib-metadata to one that's not compatible with Python 2.

Anyway, I don't understand in what case this change would be useful.  You can't use a method like this on a class with out making a class instance.

devngc:
I did not intend to change importlib-metadata. I believe it is being updated automatically.
Sure this is an instance method and is only called on a Honeybee model object. This check can be useful in a situation where a method outputs a Honeybee model and we are calling this method to write HBJSON but the method return None or an empty list instead of a Honeybee model.

mostaphaRoudsari:
> I did not intend to change importlib-metadata. I believe it is being updated automatically.

I believe your branch is rebased on the wrong branch. You should rebase to master and it will be fixed.

chriswmackey:
> This check can be useful in a situation where a method outputs a Honeybee model and we are calling this method to write HBJSON but the method return None or an empty list instead of a Honeybee model.

If this is a possibility, then I think the check should be done on whatever code is calling the `to_json` method in this unconventional way and we shouldn't pollute the core libraries with a bunch of checks that aren't needed when people are using the libraries correctly.  If we had to do a check like this for every single method that had `self` as an input, we would have so much of the code devoted to these types of checks.

Do you agree?

mostaphaRoudsari:
Hi @devngc, can you help me to understand how can you call this method before creating an instance? It should not be possible unless it is a staticmethod - which is not the case here.

devngc:
I agree with what @chriswmackey suggested. It would be better to implement this check in the library that is using Honeybee-core. I will close this PR now.

PR #366, opened 2021-02-18
fix(model): add from_hbjson method

mostaphaRoudsari:
Added a classmethod to make it easier to load hbjson files.


chriswmackey (approved):
LGTM!

PR #389, opened 2021-03-08
docs(cli): Improve CLI documentation structure

santiagogaray:
Hi @chriswmackey,

This one looks pretty straightforward, no warnings or special cases. Please let me know if you have any issue on your side!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -266,11 +266,7 @@ ht_lib_tool = {}
 #                     CLI_filename_4: group_name_4
 #                    }
 # }
-ht_cli_file_group = {
-    'honeybee_radiance': {'mtx': 'mtxop',
-                          'postprocess': 'post_process'
-                          }
-}
+ht_cli_file_group = {}
 
 
 def create_cli_files():
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,12 +14,14 @@
 #
 import os
 import sys
+import datetime
+now = datetime.datetime.now()
 sys.path.insert(0, os.path.abspath('..'))
 
 # -- Project information -----------------------------------------------------
 
 project = 'honeybee'
-copyright = '2021, Ladybug Tools'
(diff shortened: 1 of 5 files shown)


chriswmackey:
Will review soon.

docs/conf.py
```diff
@@ -19,7 +19,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'honeybee'
-copyright = '2019, Ladybug Tools'
+copyright = '2021, Ladybug Tools'
```
chriswmackey, line 22: Thank you for catching this. Mostapha actually found a great solution for this situation, which will ensure that the year is always correct, [which you can see here in honeybee-radiance](https://github.com/ladybug-tools/honeybee-radiance/blob/master/docs/conf.py#L23).  If you could implement the same here, that would be a big help.

docs/conf.py
```diff
@@ -225,6 +226,249 @@
...
+#                    },
+#    library_name_2: {CLI_filename_3: group_name_3,
+#                     CLI_filename_4: group_name_4
+#                    }
+# }
+ht_cli_file_group = {
+    'honeybee_radiance': {'mtx': 'mtxop',
+                          'postprocess': 'post_process'
+                          }
+}
```
chriswmackey, line 271: Given that this repo is does not have these `honeybee-radiance`  modules (or any modules that have a different name than the CLI group), could we remove this part?  If you want to leave the comments in case we add modules with different group names in the future, I think that's fine. But it doesn't seem right to pollute this repo with the baggage of honeybee-radiance. It's a totally separate reop, after all, and it runs independently of honeybee-radiance.

chriswmackey (commented):
Thanks @santiagogaray . It's generally looking good and I just have a couple of comments.

santiagogaray:
Thanks @chriswmackey for reviewing. Good comments, much cleaner to leave the hash table empty. Updated also the year per Mostapha's solution.

PR #512, opened 2022-03-10
Added missing user_data on face to_dict method

rvaguerra:
Method to_dict inside Face class was missing the user_data property.


chriswmackey (approved):
Thank you for catching this bug and pushing the fix, @rvaguerra ! I don't know how I missed this one.

Merged!

PR #538, opened 2022-11-07
feat(cli): Add commands for comparing and syncing models

chriswmackey:


honeybee/door.py
```diff
@@ -484,6 +484,9 @@ def is_geo_equivalent(self, door, tolerance=0.01):
         Returns:
             True if geometrically equivalent. False if not geometrically equivalent.
         """
+        tol_val = tolerance * self.area
+        if not door.area - tol_val <= self.area <= door.area + tol_val:
```
mostaphaRoudsari, line 488: Hey, @chriswmackey! Since you are using this a few times you may want to use `abs` instead.


```suggestion
        if abs(self.area - door.area) > tol_value:
```

chriswmackey: Ahh, that's much more pythonic @mostaphaRoudsari . Thank you!

